PR #6, opened 2024-01-22
Add  land check

### LINKED ISSUE #5: Divide by zero in iceberg sources
The 4P version of COBALT has the following lines [here](https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/3db75ef0f3939a85e129e98802097e45666ca337/generic_tracers/generic_COBALT.F90#L9768-L9770): 

```
cobalt%jfe_iceberg(i,j,1) = cobalt%jfe_iceberg_ratio*max(frunoff(i,j),0.0)/rho_dzt(i,j,1)
cobalt%jno3_iceberg(i,j,1) = cobalt%jno3_iceberg_ratio*max(frunoff(i,j),0.0)/rho_dzt(i,j,1)
cobalt%jpo4_iceberg(i,j,1) = cobalt%jpo4_iceberg_ratio*max(frunoff(i,j),0.0)/rho_dzt(i,j,1)

```

The problem is that `rho_dzt(:, :, 1)` (density times layer thickness) is 0 over land. 

This block probably needs to be wrapped in a check for the index of the bottom layer like  [line 9788](https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/3db75ef0f3939a85e129e98802097e45666ca337/generic_tracers/generic_COBALT.F90#L9788C8-L9788C38):
```
if (grid_kmt(i,j) .gt. 0) then
```

I don't think this affects the solution since it happens over land, but debugging other issues (like I'm currently trying to do) is harder since the model aborts here when it is compiled in debug mode.

yichengt900:
Fixed #5

### DIFF
--- a/generic_tracers/generic_COBALT.F90
+++ b/generic_tracers/generic_COBALT.F90
@@ -9764,12 +9764,14 @@ contains
     !
     ! Iceberg and "Coastal" iron and other nutrient input
     !
-    do j = jsc, jec ; do i = isc, iec
-       cobalt%jfe_iceberg(i,j,1) = cobalt%jfe_iceberg_ratio*max(frunoff(i,j),0.0)/rho_dzt(i,j,1)
-       cobalt%jno3_iceberg(i,j,1) = cobalt%jno3_iceberg_ratio*max(frunoff(i,j),0.0)/rho_dzt(i,j,1)
-       cobalt%jpo4_iceberg(i,j,1) = cobalt%jpo4_iceberg_ratio*max(frunoff(i,j),0.0)/rho_dzt(i,j,1)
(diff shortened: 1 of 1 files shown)


andrew-c-ross (approved):
Thanks for putting this together quickly! It did solve the issue that I was having.

I'm a little puzzled why this bug hasn't caused issues before (only when I switched to the NWA25 configuration), but I also don't see any reason why it would be wrong to do this check, so I'm okay with merging this.

Longer-term, there are quite a few of these checks for grid_kmt, and I wonder if that is costing us any performance and if there's a better way.

Issue #7, opened 2024-01-22
Consolidate generic tracer interface for cobalt-4p 


yichengt900:
Currently there are a few modifications are required in MOM6 generic tracer interface in order to use cobalt-4p version. They are mainly:

1. `geolonT` and `geolatT `for calculation of day-length (e.g.  https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/ec631ef65375de78c35a1891e67502b49b230507/patch/MOM6/src/tracer/MOM_generic_tracer.F90#L180)

2. `tv%eqn_of_state` for density calculation to calculate the mld for the photoacclimation (e.g. https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/3db75ef0f3939a85e129e98802097e45666ca337/generic_tracers/generic_COBALT.F90#L7856)

We want to revisit these changes and make sure they are what we want before opening a pull request to the MOM6 main branch.

PR #8, opened 2024-01-23
Simplify changes to the generic tracer interface for cobalt-p4: ocean_BGC side

yichengt900:
This PR is to address #7 . 

The [new cobalt-p4](https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC) requires several variables from MOM6 (e.g. `geolat` for day-length calculation as well as `eqs_of_state` for density calculation).

We revisited the initial modifications and removed unnecessary changes to minimize the impact on the generic tracer interface at the ocean_BGC side. The results match those obtained from the original codes in the 1-D and NWA12 RT tests. 

I will conduct a 5-year run to ensure the consistency of the results.
**Update** The restart files from this PR  (`/gpfs/f5/gfdl_med/scratch/Yi-cheng.Teng/fre/g_tracer/NWA12/NWA12_COBALT_2023_10_v2/ncrc5.intel22-prod/archive/restart`) are identical to the baselines (`/gpfs/f5/cefi/scratch/Yi-cheng.Teng/fre/cefi/NWA12/NWA12_COBALT_2023_10_v2/ncrc5.intel22-prod/archive`).

Dependency: https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/MOM6/pull/3


generic_tracers/generic_COBALT.F90
```diff
@@ -7873,11 +7871,12 @@ subroutine generic_COBALT_update_from_source(tracer_list,Temp,Salt,tv,rho_dzt,dz
     real, dimension(:,ilb:,jlb:,:), intent(in) :: opacity_band
     real, dimension(ilb:,jlb:),     intent(in) :: internal_heat
     real, dimension(ilb:,jlb:),     intent(in) :: frunoff
+    real, dimension(ilb:,jlb:),     intent(in) :: geolat
+    type(EOS_type),                 intent(in)  :: eqn_of_state !< Equation of state structure
```
theresa-cordero: should these new variables be optional here like they are in generic_tracer_source ?

generic_tracers/generic_tracer.F90
```diff
@@ -554,9 +553,9 @@ subroutine generic_tracer_source(Temp,Salt,tv,rho_dzt,dzt,hblt_depth,ilb,jlb,tau
          hblt_depth,ilb,jlb,tau,dtts,grid_dat,model_time,&
          nbands,max_wavelength_band,sw_pen_band,opacity_band, grid_ht)
 
-    if(do_generic_COBALT)  call generic_COBALT_update_from_source(tracer_list,Temp,Salt,tv,rho_dzt,dzt,&
+    if(do_generic_COBALT)  call generic_COBALT_update_from_source(tracer_list,Temp,Salt,rho_dzt,dzt,&
          hblt_depth,ilb,jlb,tau,dtts,grid_dat,model_time,&
-         nbands,max_wavelength_band,sw_pen_band,opacity_band,internal_heat,frunoff)
+         nbands,max_wavelength_band,sw_pen_band,opacity_band,internal_heat,frunoff,geolat,eqn_of_state)
```
theresa-cordero: since the geolat and eqn_of_state variables are optional in this subroutine but required for generic_COBALT_update_from_source do we need to add logic to make sure that they are not unspecified in the generic_tracer_source call but are used in this line?

theresa-cordero (commented):
This seems straightforward. I think as long as the optional arguments are setup in a way will ensure that the model is not filling things in if they are not specified.

theresa-cordero (approved):
thanks for making these changes. This looks good to me.

PR #12, opened 2024-03-08
Tentative changes from final version used for photoacclimation experiments

andrew-c-ross:
This PR would make COBALT equivalent to the version found in Charlie's generic_COBALT_08222023_control.F90. Some of the changes in this PR are specific to the needs of the global photoacclimation runs, however, and shouldn't be merged. Reviews would be appreciated to identify what should or shouldn't be merged.

### DIFF
--- a/generic_tracers/generic_COBALT.F90
+++ b/generic_tracers/generic_COBALT.F90
@@ -248,8 +248,8 @@ namelist /generic_COBALT_nml/ do_14c, co2_calc, debug, do_nh3_atm_ocean_exchange
           si_2_n_max,    &
           si_2_n_static, &
           thetamax,      &
-          bresp_frac_mixed, &
-          bresp_frac_strat, &
+          bresp_frac_norm, &
+          bresp_frac_stress, &
           sink_max,      &
           agg,           &
           frac_mu_stress, &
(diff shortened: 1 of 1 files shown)


generic_tracers/generic_COBALT.F90
```diff
@@ -6341,14 +6341,14 @@ subroutine user_add_params
...
-    call g_tracer_add_param('p_2_n_min_Lg', phyto(LARGE)%p_2_n_min,1.0/31.0 )               ! mol P mol N-1
-    call g_tracer_add_param('p_2_n_slope_Lg', phyto(LARGE)%p_2_n_slope, 0.048*1.0e6)        ! mol P mol N-1 mol P-1 kg
+    call g_tracer_add_param('p_2_n_min_Sm', phyto(SMALL)%p_2_n_min,1.0/23.0 )               ! mol P mol N-1
+    call g_tracer_add_param('p_2_n_slope_Sm', phyto(SMALL)%p_2_n_slope, 0.0*1.0e6)        ! mol P mol N-1 mol P-1 kg 
+    call g_tracer_add_param('p_2_n_max_Sm', phyto(SMALL)%p_2_n_max,1.0/23.0 )               ! mol P mol N-1
+    call g_tracer_add_param('p_2_n_min_Md', phyto(MEDIUM)%p_2_n_min,1.0/18.0 )              ! mol P mol N-1
+    call g_tracer_add_param('p_2_n_slope_Md', phyto(MEDIUM)%p_2_n_slope, 0.0*1.0e6)       ! mol P mol N-1 mol P-1 kg
+    call g_tracer_add_param('p_2_n_max_Md', phyto(MEDIUM)%p_2_n_max,1.0/18.0 )              ! mol P mol N-1
+    call g_tracer_add_param('p_2_n_min_Lg', phyto(LARGE)%p_2_n_min,1.0/14.0 )               ! mol P mol N-1
+    call g_tracer_add_param('p_2_n_slope_Lg', phyto(LARGE)%p_2_n_slope, 0.0*1.0e6)        ! mol P mol N-1 mol P-1 kg
```
andrew-c-ross, line 6351: I think I see what's going on here now... this is effectively disabling the variable n:p, right? So this should not be merged.

yichengt900:
Thanks @andrew-c-ross! @charliestock, based on @nikizadehgfdl's comments during the `COBALTv3 Focused Doc and Dev` meeting, it seems that we can overwrite parameters using field_table without recompiling the code. This raises the following question: do we still want to update the default values to match those used in the manuscript, purely for the sake of consistency and reference?

yichengt900:
@charliestock, following our discussion, I think we will eventually go down the route of providing users with our recommended BGC parameters through a separate file, rather than editing the default values in the code.  If so, I will close this PR for now and keep it as a reference. @andrew-c-ross are you ok with that?

andrew-c-ross:
I think that sounds ok. Yeah I will keep my branch around, so with that branch and this closed PR we can always compare values if needed.

Issue #14, opened 2024-03-19
Cleanup idea: eliminate repeated diagnostic ID checks


andrew-c-ross:
For generic tracer diagnostic writing, every time before data is posted, the code checks if the diagnostic ID is greater than 0 (if the diagnostic is being used). For example,

https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/bfa02118f1bda46cc230d18074401e6d782a8e74/generic_tracers/generic_COBALT.F90#L7586-L7589

I'm pretty sure this check could be moved to within the g_send_data_* functions within generic_tracer_utils.F90. Doing so would allow the elimination of over 600 lines of `if (cobalt%id_* .gt. 0)` code from COBALT.

yichengt900:
@andrew-c-ross, thanks for the great cleanup ideas (including #15)! I believe both ideas are pretty straightforward and easy to implement. I could do both of them on my end, but I would love to use this opportunity to exercise our code review process and use this as an excellent example for other group members.

If you are okay with that, I am going to add a `cleanup` label and assign this issue to you, and you could open a PR associated with your cleanup idea. To save your time, you just need to fix one tracer diagnostic writing, and I can implement the others. By doing this, we will have a clean example for other developers.

@charliestock, do you think this is a good idea? Any thoughts are more than welcome!

andrew-c-ross:
Sounds great, @yichengt900. Would also be good to get the opinion and advice of @nikizadehgfdl before we go about deleting all of these lines.

yichengt900:
Good point, @andrew-c-ross. @nikizadehgfdl, @andrew-c-ross has proposed two great ideas (this issue and #15) to eliminate repeated code in COBALT. We would love to hear your opinions before moving forward. Any inputs are more than welcome!

yichengt900:
The modified g_sed_data function has been merged through PR #16. Now we can consider implementing this new function to eliminate repeated ID check lines..  

 Also inspired by @nikizadehgfdl 's [idea](https://github.com/NOAA-GFDL/ocean_BGC/pull/31/files), we can further consider putting all send_diag calls in a separate subroutine to make the `update_from_source` subroutine easier to read. This change will also make our lives easier if we eventually decide to go with the route of semi-auto generation of diagnostic code (e.g., from a YAML file).
 
[...]

PR #16, opened 2024-03-20
Check if diagnostic id is > 0 in g_send_data_* functions - Part1

andrew-c-ross:
This adds a check for whether the diagnostic field ID is greater than zero to the g_send_data functions. If the ID is not greater than zero, sending data does nothing and now explicitly returns False. If the diagnostic is used, it behaves as usual and shouldn't change answers.

The check is only added to the MOM6 method for posting diagnostic data, since the FMS method already has a check built in ([here](https://github.com/NOAA-GFDL/FMS/blob/1bb706cc61f8351f176e4f9b7ccc8ab25ed6999b/diag_manager/diag_manager.F90#L1533)).

This is the first step to solving #14. With this in place, all of the checks for the diagnostic id can be removed from the generic tracer packages.


yichengt900:
@andrew-c-ross, thank you so much for your prompt action! I really appreciate it. The changes look great to me. I will try to invite @nikizadehgfdl as a reviewer too and hope that he is willing to provide his valuable opinions.

yichengt900:
Hey @andrew-c-ross, I've made a slight modification to your title to reflect that we'll be addressing #14 in two steps. I hope that's okay with you. I'll approve part 1 shortly, and then we can start working on implementing this new function.

yichengt900 (approved):
Codes look good. Approved.

yichengt900:
@andrew-c-ross, could you merge the recent change in dev/cefi (#19) into your branch? Just want to ensure your feature branch is up-to-date.

Issue #18, opened 2024-03-25
Missing Pointers for Benthic Fluxes in Direct Sinking of Phytoplankton P


yichengt900:
In the recent code review, we discovered that the pointers for the benthic fluxes for the direct sinking of phytoplankton P were not correctly set in the `generic_COBALT_update_from_bottom` subroutine. This may be one reason why our P budget analysis cannot be closed. Additionally, this discovery motivates us to reconsider our current imbalance check approach, as it may not detect issues occurring outside the COBALT modules.

This issue is being raised on behalf of @charliestock.

charliestock:
Thank you for handling this Yi-Cheng,  The net effect of this was a small amount of unintended phosphate burial.  This issue only impacts COBALTv3.0 and does not effect past versions of COBALT (for which phytoplankton sinking was not specified).  Furthermore, the unintended P burial was fortunately very small and did not create areas of P limitation.

PR #19, opened 2024-03-25
+Missing Pointers for Benthic Fluxes in Direct Sinking of Phytoplankton P 

yichengt900:
As @charliestock suggested, now the pointers for the benthic fluxes for the direct sinking of phytoplankton P are set correctly in `generic_COBALT_update_from_bottom` subroutine. This PR is to solve #18. 

@charliestock , While reviewing the `generic_COBALT_update_from_bottom` subroutine, I noticed that pointers for the benthic fluxes for the direct sinking of phytoplankton Fe are also turned off. Below is an example for diazotroph Fe:

https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/01f7a0cff3270d6ef076be301b9417a0c1e3f83f/generic_tracers/generic_COBALT.F90#L7707-L7708 

Could you confirm if this was done intentionally?

### DIFF
--- a/generic_tracers/generic_COBALT.F90
+++ b/generic_tracers/generic_COBALT.F90
@@ -7746,6 +7746,8 @@ contains
     !
     call g_tracer_get_values(tracer_list,'pdi','btm_reservoir',phyto(DIAZO)%fp_btm,isd,jsd)
     phyto(DIAZO)%fp_btm = phyto(DIAZO)%fp_btm/dt
+    call g_tracer_get_pointer(tracer_list,'pdi_btf','field',temp_field)
+    temp_field(:,:,1) = phyto(DIAZO)%fp_btm(:,:)
     call g_tracer_set_values(tracer_list,'pdi','btm_reservoir',0.0)
     if (phyto(DIAZO)%id_fp_btm .gt. 0)           &
          used = g_send_data(phyto(DIAZO)%id_fp_btm,phyto(DIAZO)%fp_btm, &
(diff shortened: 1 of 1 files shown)


charliestock (approved):
Thanks for implementing this Yi-Cheng.  It looks good to me.  If we could run just confirm conservation in the 1D case, that would be great.

charliestock:
Thanks Yi-Cheng,  In answer to your question about the iron, this was indeed intentional.  The iron flux to the sediment is removed.  The flux from the sediment is then set using a published relationship:

https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/bfa02118f1bda46cc230d18074401e6d782a8e74/generic_tracers/generic_COBALT.F90#L9903-L9906

yichengt900:
> Thanks Yi-Cheng, In answer to your question about the iron, this was indeed intentional. The iron flux to the sediment is removed. The flux from the sediment is then set using a published relationship:
> 
> https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/bfa02118f1bda46cc230d18074401e6d782a8e74/generic_tracers/generic_COBALT.F90#L9903-L9906

Thanks @charliestock .  I believe the relationship is from [Dale et al. 2015](https://agupubs.onlinelibrary.wiley.com/doi/full/10.1002/2014GB005017), is that correct? I will add more detailed inline documentation to highlight this.

yichengt900:
> Thanks for implementing this Yi-Cheng. It looks good to me. If we could run just confirm conservation in the 1D case, that would be great.

@charliestock . Thanks. I can confirm with this fix, the total P is conservative in the 1D case by checking `wc_vert_int_p`: 
 
Screenshot from 2024-03-26 12-59-53

yichengt900:
This PR passed the CI test and has proven mass conservation for P in the 1D case. Merge now.

Issue #20, opened 2024-03-26
Anomalous namelist usage early in the code to set parameters


charliestock:
There are a small number of parameters being defined in a name  list at the beginning of the code:

https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/bfa02118f1bda46cc230d18074401e6d782a8e74/generic_tracers/generic_COBALT.F90#L197-L227

Many of these are a legacy from Fabien Paulot's development of dynamic ammonia exchanges across the air-sea interface and can likely be either removed and/or set as the other parameters are set.  I can put together a pull request to remove these once we agree on the best approach for handling the setting of parameter values.

yichengt900:
@charliestock , thanks for bringing this up. I also noticed that during my code cleanup & inline doc process (#17). At this moment I simply commented out those lines in  branch `feature/inline_documentation`:

https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/9fceed8e0cf033f2f09e7030fb2748ce05594af7/generic_tracers/generic_COBALT.F90#L204-L212

But we definitely should remove them if they are not used anymore. 

Regarding how to handle the setting of parameter values, I like the `g_tracer_add_param` way because we can always overwrite default values through field_table, but I am open to any ideas.

yichengt900:
Some of the unused parameters defined in a namelist have been removed in PR #23. The remaining parameters have now been moved to `cobalt_types.F90`:

```
  character(len=10), public ::  co2_calc = 'mocsy'           !< carbonate formalation options. Default is 'mocsy'
  logical, public :: do_14c             = .false.            !< If true, then simulate radiocarbon 
  logical, public :: do_nh3_atm_ocean_exchange = .false.     ! If true, then do NH3 air-sea exchange 
  !
  logical, public :: do_vertfill_pre = .false.             !< Returns tracer arrays with sensible values
  logical, public :: debug           = .false.             !< not use   
  real, public    :: o2_min_nit= 0.01e-6                   !< Oxygen threshold for nitrification (mol O2 kg-1)
  real, public    :: k_o2_nit  = 3.9e-6                    !< Oxygen half saturation constant for nitrification
  real, public    :: irr_inhibit = 10.                     !< Irradiance inhibition term for nitrification (W m-2)
  real, public    :: gamma_nitrif= 3.5e6                   !< Rate constant for nitrification (month-1)
[...]

PR #23, opened 2024-03-29
Eliminate repeated diagnostic ID checks and refactor send_diag section

yichengt900:
This PR addresses issue #14. Several code changes are included:

1. Removed diagnostic ID check lines from the original `generic_COBALT_update_from_source`  and `generic_COBALT_update_from_bottom` subroutines .
2. Created a new file `cobalt_glbl.F90` and moved `phytoplankton`, `zooplankton`, `bacteria`, and `generic_cobalt` data types as well as `namelist` here.
4. Created a new file `cobalt_send_diag.F90` and added a new subroutine `cobalt_send_diagnostics` to the `cobalt_send_diag` module to handle send_diag calls separately.
5. Added a new subroutine `cobalt_send_diagnostics_bottom` to the `cobalt_send_diag` module to handle send_diag calls for bottom fluxes.
6. Reduced the total number of lines in `generic_COBALT.F90` from `16169` to `11484` (a `28.91`% reduction).
7. Regarding computational time, the average runtime before and after refactoring in 1d case is `11.50` seconds versus `11.60` seconds, representing an approximate `0.87`% increase.
8. The following namelist capabilities have been removed (mentioned by issue #20):
```console
k_nh4_small
k_nh4_diazo
k_nh4_large
k_no3_small
k_no3_diazo
```
I'm still keeping the `do_14c` option and all the diagnostics related to radiocarbon. We can remove them later if we decide to go in that direction.

This PR should not change baselines (passed CI test) and can produce b2b diagnostic netCDF files.

Eventually we could apply this refactoring to the majority of existing generic_COBALT subroutines (e.g., PR #22 ), which will enhance the clarity and conciseness of the main COBALT code.


generic_tracers/cobalt_glbl.F90
```diff
@@ -0,0 +1,1531 @@
...
+  implicit none
+  public  
+
+  !The following variables for using this module
+  ! are overwritten by generic_tracer_nml namelist
+  logical :: do_nh3_diag                             !< logic for setting NH3 diagnostic tracer field 
+
+! Namelist Options
+
+  character(len=10) ::  co2_calc = 'ocmip2'          !< carbonate formalation options. Other option is 'mocsy'
```
andrew-c-ross: This could be a separate PR, but now might be a good time to have this default to "mocsy".

generic_tracers/cobalt_glbl.F90
```diff
@@ -0,0 +1,1531 @@
...
+  integer :: scheme_no3_nh4_lim = 2 !< Nitrate and ammonia limitation scheme options
+                                    !! 1-Frost and Franzen (1992)
+                                    !! 2-O'Neill
+
+  integer :: scheme_nitrif = 3      !< nitrification scheme options:
+                                    !! 1-default COBALT
+                                    !! 2-update with no temperature dependence
+                                    !! 3-update with temperature dependence
+
+namelist /generic_COBALT_nml/ do_14c, co2_calc, do_nh3_atm_ocean_exchange, scheme_nitrif, debug, &
```
andrew-c-ross: This won't affect the bug with the air-sea flux setting. I can send a separate PR on top of this one to finally resolve that.

generic_tracers/cobalt_glbl.F90
```diff
@@ -0,0 +1,1531 @@
+!> COBALT_glbl module consists of core parameters and vars 
+!! to be used by generic COBALT related modules
+!<----------------------------------------------------------------
+module cobalt_glbl
```
charliestock: Following our conversation yesterday, do we want to call this module "cobalt_types"?

generic_tracers/cobalt_glbl.F90
charliestock: Based on John's comments, I think this would be safe change to make and could be made now.  

There was an comment about wanting to remove "ocmip" from the name of the old routine (and the routine itself), but perhaps we can address that in a different PR.

generic_tracers/cobalt_glbl.F90
```diff
@@ -0,0 +1,1531 @@
...
+
+! Namelist Options
+
+  character(len=10) ::  co2_calc = 'ocmip2'          !< carbonate formalation options. Other option is 'mocsy'
+  logical :: do_14c             = .false.            !< If true, then simulate radiocarbon 
+  logical :: do_nh3_atm_ocean_exchange = .false.     ! If true, then do NH3 air-sea exchange 
+  !
+  logical :: do_vertfill_pre = .false.             !< Returns tracer arrays with sensible values
+  logical :: debug              = .false.          !< not use   
+  real    :: o2_min_nit= 0.01e-6                   !< Oxygen threshold for nitrification (mol O2 kg-1)
```
charliestock: We could enter the values for o2_min_nit, k_o2_nit, irr_inhibit, gamma_nitrif and k_nh4_nitrif as we do the other model parameters as part of the namelist cleanup (i.e., in user_add_params).

generic_tracers/cobalt_send_diag.F90
```diff
@@ -0,0 +1,2622 @@
...
+        model_time, rmask = grid_tmask(:,:,1),&
+        is_in=isc, js_in=jsc, ie_in=iec, je_in=jec)
+
+!==============================================================================================================
+
+    end subroutine cobalt_send_diagnostics   
+
+
+    !> subroutine that handles send_diag calls for bottom fluxes      
+    subroutine cobalt_send_diagnostics_bottom(tracer_list,model_time,grid_tmask,temp_field,dt,&
```
charliestock: This routine does more than just send the bottom diagnostics.  It also updates the bottom reservoirs and sets the bottom fluxes for later application as a bottom boundary condition.  I think the latter function may be more important than its diagnostic role (and was captured by the "update_from_bottom" naming.  

We may be able to separate the diagnostic function from "update_from_bottom", but I do wonder whether the timing of the bottom update relative to "update_from_source" may be a consideration here.  Ideally, the bottom flux save would be aligned with the time step during which it is applied in the tridiagonal solver.  We should be able to determine this once we have a flow diagram of generic tracer interactions with MOM6.

For now, maybe we should keep the generic_cobalt_update_from_bottom intact for this pull request and open and open issue on the bottom diagnostics that we can resolve later?

generic_tracers/cobalt_glbl.F90
```diff
@@ -0,0 +1,1531 @@
+!> COBALT_glbl module consists of core parameters and vars 
+!! to be used by generic COBALT related modules
+!<----------------------------------------------------------------
+module cobalt_glbl
+  use field_manager_mod, only: fm_string_len, fm_path_name_len      
+  implicit none
+  public
```
theresa-cordero: The MOM6 convention would be to use 
```
implicit none ; private 
public [list of public types/routines in this module]
```
here.  I think this would impact the parameters listed from lines ~11-54 which are not within a type. But I think each one could be specified as public.
I personally like this format because it is a quick way to find out what types/routines are within a file when you first open it.

generic_tracers/generic_COBALT.F90
```diff
@@ -7575,232 +6046,8 @@ subroutine generic_COBALT_update_from_bottom(tracer_list, dt, tau, model_time)
...
-    call g_tracer_get_pointer(tracer_list,'simd_btf','field',temp_field)
-    temp_field(:,:,1) = phyto(MEDIUM)%fsi_btm(:,:)
-    call g_tracer_set_values(tracer_list,'simd','btm_reservoir',0.0)
-    if (phyto(MEDIUM)%id_fsi_btm .gt. 0)           &
-         used = g_send_data(phyto(MEDIUM)%id_fsi_btm,phyto(MEDIUM)%fsi_btm, &
-         model_time, rmask = grid_tmask(:,:,1),&
-         is_in=isc, js_in=jsc,ie_in=iec, je_in=jec)
-
+    call cobalt_send_diagnostics_bottom(tracer_list,model_time,grid_tmask, &
+                            temp_field,dt,isc,iec,jsc,jec,isd,jsd,phyto,cobalt)
```
charliestock: See comment above concerning the role of this routine in updating the bottom reservoirs and setting the bottom fluxes.  Maybe keep as "update_from_bottom" for now?

generic_tracers/cobalt_glbl.F90
yichengt900: Change `co2_calc = 'mocsy'`

generic_tracers/cobalt_glbl.F90
yichengt900: rename `cobalt_glbl` to `cobalt_types`

generic_tracers/cobalt_glbl.F90
yichengt900: Agreed. I believe we should maintain a consistent approach for reading all the model input parameters. I will retain those parameters for now, and we should discuss our preference regarding how to handle COBALT parameter inputs.

generic_tracers/cobalt_send_diag.F90
yichengt900: Agreed. I have reverted these changes, and we will revisit the bottom flux issue later.

generic_tracers/cobalt_glbl.F90
yichengt900: Make the module private now and specify individual variables as public.

generic_tracers/cobalt_glbl.F90
yichengt900: Sounds good. I will keep this one and it will fix by another PR.

yichengt900:
This PR addresses all the comments from reviewers and is up-to-date with the main branch. Please review it again, and feel free to provide any further comments, thanks.

charliestock:
Thanks for all of your hard work on the Yi-Cheng, all.  It looks good from my end.

yichengt900:
Thanks, @charliestock. I've also merged #39 into this PR. It's ready for merging once one of you approves it.

charliestock (approved):
This looks good to me Yi-Cheng.  Why don't we move forward with merging and discuss further at the doc and dev tomorrow.

PR #31, opened 2024-04-04
added felim to diazo iron limitation

charliestock:
This change updated the diazotroph iron limitation to be consistent with a change in the iron limitation calculation made for the other phytoplankton types in COBALTv3.0.  The level of iron limitation is now calculated as the maximum of the limitations imposed by the internal cell quota (def_fe) and the ambient iron uptake limitation (felim).  

While this fix changes answers, the effect is very small.  It mainly removes small regions near river mouth where iron ends up being the most limiting nutrient simply because it takes a short time for high internal iron quotas to be established in high iron environments.   Using the maximum of def_fe and felim removes this delay and puts iron on equal footing with N and P.

This adjustment also only effects the diazotrophs, which are generally a small fraction of the phytoplankton community.

### DIFF
--- a/generic_tracers/generic_COBALT.F90
+++ b/generic_tracers/generic_COBALT.F90
@@ -8276,7 +8276,7 @@ contains
     do k = 1, nk ; do j = jsc, jec ; do i = isc, iec   !{
        n=DIAZO
        phyto(n)%liebig_lim(i,j,k) = phyto(n)%o2lim(i,j,k)* &
-          min(phyto(n)%po4lim(i,j,k), phyto(n)%def_fe(i,j,k))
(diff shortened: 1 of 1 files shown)


yichengt900:
Thanks, @charliestock. It looks good to me. As you pointed out, the effect of this change is probably small and was not even caught by our 1D test case (which makes sense because our 1D test case is located in the middle of the ocean). I will approve it for now and will merge this PR once it has passed the NWA12 RT test.

yichengt900 (approved):

yichengt900:
Ok it also passed the [NWA12 RT test](https://github.com/NOAA-GFDL/CEFI-regional-MOM6/actions/runs/8558281033/job/23453836684?pr=14) at the CEFI-regional-mom6 side. Will merge this PR now.

Issue #29, opened 2024-04-04
Getting an error from the equation of state, called from COBALT


kshedstrom:
Yesterday's dev/cefi gave me this stack trace:
```
fms_MOM6_SIS2_COB 00000000009560C2 mom_eos_mp_calcul     228 MOM_EOS.F90
fms_MOM6_SIS2_COB 0000000001727663 generic_cobalt_mp    8320 generic_COBALT.F90
fms_MOM6_SIS2_COB 0000000001439637 generic_tracer_mp     563 generic_tracer.F90
fms_MOM6_SIS2_COB 0000000001351345 mom_generic_trace     582 MOM_generic_tracer.F90
fms_MOM6_SIS2_COB 0000000000FD1A8F mom_tracer_flow_c     529 MOM_tracer_flow_control.F90
fms_MOM6_SIS2_COB 0000000000A2F3E3 mom_diabatic_driv    1607 MOM_diabatic_driver.F90
fms_MOM6_SIS2_COB 0000000000A24105 mom_diabatic_driv     404 MOM_diabatic_driver.F90
fms_MOM6_SIS2_COB 00000000009D2020 mom_mp_step_mom_t    1602 MOM.F90
fms_MOM6_SIS2_COB 00000000009C78FB mom_mp_step_mom_     851 MOM.F90
fms_MOM6_SIS2_COB 00000000009751C2 ocean_model_mod_m     594 ocean_model_MOM.F90
fms_MOM6_SIS2_COB 0000000000413DFF MAIN__          1063 coupler_main.F90
```
I recompiled with ec631ef65375de78c35a1891e67502b49b230507 and it runs past this (and runs into the imbalance issues).

andrew-c-ross:
The depth index for the temperature and salinity used to calculate density, `kmld_ref`, is a float rather than an int. I think this was also present in ec631ef, so maybe it's not the problem, but just wanted to point that out.

https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/6e6ed114a711f060da78aece6170a4369b2ee32c/generic_tracers/generic_COBALT.F90#L7915

https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/6e6ed114a711f060da78aece6170a4369b2ee32c/generic_tracers/generic_COBALT.F90#L8320

yichengt900:
[...]

PR #32, opened 2024-04-04
Cleanup: float index and unused variables

andrew-c-ross:
I started this PR to fix the extremely minor issue of using a float as an index (mentioned in #29). While I was at it, though, I also cleaned up some other extremely minor issues reported by compiling with -Wall (some unused variables and all instances of tabs). None of these should have any effect on the actual model. There's a chance the unused variables indicate something that is unintentionally not included in the current source, though.

### DIFF
--- a/generic_tracers/generic_COBALT.F90
+++ b/generic_tracers/generic_COBALT.F90
@@ -7877,7 +7877,7 @@ contains
     type(EOS_type),             optional, intent(in) :: eqn_of_state !< Equation of state structure
 
     character(len=fm_string_len), parameter :: sub_name = 'generic_COBALT_update_from_source'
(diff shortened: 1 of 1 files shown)


yichengt900:
Thanks, @andrew-c-ross. Although some of the changes, like tab removal, might be overwritten by #23, they all look good to me. I will keep this PR open for a while to give others a chance to review it

andrew-c-ross:
Sounds good :+1: It's all extremely picky details that won't change anything, so there's no urgency

yichengt900 (approved):

charliestock:
Thanks for catching these Andrew, Yi-Cheng.  It is good to catch these whether they are small or not!

PR #42, opened 2024-04-09
Fix setting of as_param and allow generic_COBALT to set as_param_COBALT

### LINKED ISSUE #9: Air-sea parameter inconsistencies and bugs
There are some long-standing issues with the air-sea flux parameter settings that might be worth fixing while we're making breaking changes to the generic tracers.

Most of the generic tracer packages have their own air-sea parameter setting as a module variable. For example, for COBALT,
https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/4af91db7215e07fbf6c537fd298ad66a72da0f14/generic_tracers/generic_COBALT.F90#L185

These parameter settings are intended to be set to a common value by setting `as_param` in the generic tracer namelist:
https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/4af91db7215e07fbf6c537fd298ad66a72da0f14/generic_tracers/generic_tracer.F90#L144-L148

However, the module variables for each tracer package are only set if `as_param == 'gfdl_cmip6'`:
https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/4af91db7215e07fbf6c537fd298ad66a72da0f14/generic_tracers/generic_tracer.F90#L173-L179

If `as_param` is anything other than "gfdl_cmip6" (its default value), then each generic tracer uses whatever its own default as_param is. This means if you set, for example, `as_param = W14`, a package like BLING will not see that and use its default "gfdl_cmip6" instead. 

In the NWA12 XML, we set `as_param = unused` so that `as_param_cobalt` won't be set and can use its default value of W92, which works but is unpleasant.

I think this could be fixed while retaining compatibility with old global solutions (but requiring changes to regional XMLs) by just deleting:
https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/4af91db7215e07fbf6c537fd298ad66a72da0f14/generic_tracers/generic_tracer.F90#L173

But this leaves a bunch of module variables in place that don't do anything, so it would be nice if we could have either one single air-sea param or else each generic tracer check its own namelist.

andrew-c-ross:
This fixes #9 by always setting the air-sea parameters for each generic tracer package to the value set by as_param in generic_tracer_nml. For COBALT, it also allows overriding the globally set as_param by setting as_param_cobalt in generic_COBALT_nml. (This probably won't be used, but the option is there since each generic tracer package is maintaining its own as_param).

I'd appreciate a double-check that this works correctly in case generic_COBALT.F90 is not always setting the value of the as_param_cobalt module variable before generic_COBALT.F90 reads and overwrites from the namelist.

### DIFF
--- a/generic_tracers/generic_tracer.F90
+++ b/generic_tracers/generic_tracer.F90
@@ -141,7 +141,7 @@ module generic_tracer
   logical :: do_vertfill_post = .false.
   logical :: generic_tracer_register_called = .false.
   logical :: force_update_fluxes = .false.
-  character(len=10) :: as_param   = 'gfdl_cmip6'     ! Use default Wanninkhoff/OCMIP2 parameters for air-sea gas transfer
+  character(len=10) :: as_param   = 'W14'     ! Use Wanninkhoff 2014 parameters for air-sea gas transfer by default
 
   namelist /generic_tracer_nml/ do_generic_tracer, do_generic_abiotic, do_generic_age, do_generic_argon, do_generic_CFC, &
       do_generic_SF6, do_generic_TOPAZ,do_generic_ERGOM, do_generic_BLING, do_generic_miniBLING, do_generic_COBALT, &
@@ -169,7 +169,6 @@ contains
     write (stdoutunit, generic_tracer_nml)
     write (stdlogunit, generic_tracer_nml)
 
-    ! Use Wanninkhoff 2014 parameters for air-sea gas exchange if as_param='W14' in generic_tracer_nml
     ! The air-sea parameters for each generic tracer package default
     ! to being the same as as_param.
     ! For generic_COBALT, the as_param can be overwritten by setting
--- a/generic_tracers/generic_tracer.F90
+++ b/generic_tracers/generic_tracer.F90
@@ -170,13 +170,15 @@ contains
     write (stdlogunit, generic_tracer_nml)
 
     ! Use Wanninkhoff 2014 parameters for air-sea gas exchange if as_param='W14' in generic_tracer_nml
-    if (as_param == 'gfdl_cmip6') then
-      if (do_generic_abiotic) as_param_abiotic = as_param
-      if (do_generic_CFC)     as_param_cfc     = as_param
-      if (do_generic_SF6)     as_param_sf6     = as_param
-      if (do_generic_BLING)   as_param_bling   = as_param
-      if (do_generic_COBALT)  as_param_cobalt  = as_param
-    endif
+    ! The air-sea parameters for each generic tracer package default
+    ! to being the same as as_param.
+    ! For generic_COBALT, the as_param can be overwritten by setting
+    ! as_param_cobalt in generic_COBALT_nml.
+    if (do_generic_abiotic) as_param_abiotic = as_param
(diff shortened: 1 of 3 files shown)


andrew-c-ross:
Also note our BGC xmls, which have been using `as_param = 'unused'` as a way of being able to set `as_param_cobalt`, will need to have this line removed.

generic_tracers/generic_tracer.F90
```diff
@@ -141,7 +141,7 @@ module generic_tracer
   logical :: do_vertfill_post = .false.
   logical :: generic_tracer_register_called = .false.
   logical :: force_update_fluxes = .false.
-  character(len=10) :: as_param   = 'gfdl_cmip6'     ! Use default Wanninkhoff/OCMIP2 parameters for air-sea gas transfer
+  character(len=10) :: as_param   = 'W14'     ! Use Wanninkhoff 2014 parameters for air-sea gas transfer by default
```
yichengt900, line 144: @andrew-c-ross, everything looks good to me. I'll test it locally to ensure it works as expected. Quick question: since each generic tracer package maintains its own `as_param`, changing the default value here shouldn't affect other applications like global simulations, correct?

yichengt900:
Ok, I can confirm now that if we remove `as_param` from input.nml directly, it will take the default value `W14` as defined in generic_tracer.F90, and the results will be the same as if you had set as_param='W14' directly in input.nml. Similarly, you will get the B2B results (compared to the current version) if you set as_param='W92' in input.nml. This should act as expected.

generic_tracers/generic_tracer.F90
andrew-c-ross: That's a good point I forgot, changing the default here will actually change the default for every generic tracer that uses the air-sea parameters. I think this is desirable, and what @charliestock suggested yesterday, since the old default gfdl_cmip6 uses a mix of W92 and W14. Setting the namelist `as_param = 'gfdl_cmip6'` would be needed to reproduce old answers.

andrew-c-ross: (Each tracer package has its own as_param, but before the fixes in this PR there was no way for them to be anything other than 'gfdl_cmip6', even if you changed as_param in the namelist)

yichengt900:
Thank you for confirming, @andrew-c-ross. I have tested this PR, and it passed both the 1D and NWA12 RT cases locally. 

To reproduce the old answers, I had to set the namelist `as_param to 'W92'` in the [test cases](https://github.com/NOAA-GFDL/CEFI-regional-MOM6/blob/deed7164417c0ed67a9685611541a929fa634aad/exps/OM4.single_column.COBALT/input.nml#L138). 

@andrew-c-ross, would you mind changing the [default branch](https://github.com/andrew-c-ross/CEFI_ocean_BGC/blob/8e437d4168a628d2b8b5e625ad2f7e3c66a612d9/.github/workflows/cobalt_ci.yml#L14) in our GitHub Actions workflow yaml from `main` to `develop`? This change should ensure that this PR passes the checks.

yichengt900 (approved):

PR #43, opened 2024-04-10
New module for diagnostic registration

yichengt900:
This PR is the first step in addressing issue #15. We added a new module, `COBALT_reg_diag`, which contains the subroutine `cobalt_reg_diagnostics` to handle diagnostic registration. This reduces the total line count in `generic_COBALT.F90` from `11687` to `7485`.

This PR does not change baselines; all the diag outputs are identical to the ones from the current version.

Next, we can consider adopting Kelly's approach #22 to automatically create the `COBALT_reg_diag` module.


theresa-cordero (approved):
This looks great!

Issue #21, opened 2024-03-27
cleanup: please consider deleting some old modules


nikizadehgfdl:
Some of the generic modules are very old and have not been tested for a long time and probably won't work with MOM6 anymore. They just add to the compilation process load and time.
These include
generic_ERGOM.F90
generic_miniBLING.F90
generic_TOPAZ.F90

Please consider removing them.

yichengt900:
@charliestock, as @nikizadehgfdl suggested, do you think it's safe to remove ERGOM, miniBLING, and TOPAZ since no one uses them anymore?

PR #44, opened 2024-04-11
Remove ERGOM, TOPAZ, and miniBLING modules

yichengt900:
As @nikizadehgfdl suggested in #21, the pull request (PR) removes the `ERGOM`, `TOPAZ`, and `miniBLING` modules, as well as associated variables and calls, from `generic_tracer.F90`.


charliestock:
Following discussion with Yi-Cheng, this should be OK as long as we have other module options as a template (e.g., BLING, CFCs), and then lets rename the .F90, .F90.archive for TOPAZ, minibling and ERGOM, and then add a README.  This should prevent compiling, and we can remove more completely if these modules continue to be unused.

yichengt900:
So instead of removing these modules directly, we rename them as `generic_*.F90.archive`, and we commented out all the related calls in `generic_tracer.F90` for now.

generic_tracers/generic_tracer.F90
```diff
@@ -262,8 +262,8 @@ subroutine generic_tracer_init(isc,iec,jsc,jec,isd,ied,jsd,jed,nk,ntau,axes,grid
 
     !Allocate and initialize all registered generic tracers
     !JGJ 2013/05/31  merged COBALT into siena_201303
-    if(do_generic_abiotic .or. do_generic_age .or. do_generic_argon .or. do_generic_CFC .or. do_generic_SF6 .or. do_generic_TOPAZ &
-       .or. do_generic_ERGOM .or. do_generic_BLING .or. do_generic_miniBLING .or. do_generic_COBALT .or. do_generic_blres) then
+    if(do_generic_abiotic .or. do_generic_age .or. do_generic_argon .or. do_generic_CFC .or. do_generic_SF6  &
+       .or. do_generic_BLING .or. do_generic_COBALT .or. do_generic_blres) then
```
andrew-c-ross: Here the parts for TOPAZ, ERGOM, and miniBLING were deleted rather than commented out, so undoing this PR would not be as simple as un-commenting. To make it possible to revert while also simplifying the code, can we replace this whole statement with `if(do_generic_tracer) then`? The initialization currently gets called for every tracer that is used, and it seems like any other generic tracer that would be added would also need to initialize.

generic_tracers/generic_tracer.F90
```diff
@@ -645,8 +645,8 @@ subroutine generic_tracer_vertdiff_G(h_old, ea, eb, dt, kg_m2_to_H, m_to_H, tau)
 
     !nnz: Should I loop here or inside the sub g_tracer_vertdiff ?    
     !JGJ 2013/05/31  merged COBALT into siena_201303
-    if(do_generic_abiotic .or. do_generic_age .or. do_generic_argon .or. do_generic_CFC .or. do_generic_SF6 .or. do_generic_TOPAZ &
-       .or. do_generic_ERGOM .or. do_generic_BLING .or. do_generic_miniBLING .or. do_generic_COBALT .or. do_generic_blres) then
+    if(do_generic_abiotic .or. do_generic_age .or. do_generic_argon .or. do_generic_CFC .or. do_generic_SF6  &
+       .or. do_generic_BLING .or. do_generic_COBALT .or. do_generic_blres) then
```
andrew-c-ross: Also replace this with `if(do_generic_tracer) then`?

generic_tracers/generic_tracer.F90
```diff
@@ -687,8 +687,8 @@ subroutine generic_tracer_vertdiff_M(dh, dhw, diff_cbt, dt, Rho_0,tau)
 
     !nnz: Should I loop here or inside the sub g_tracer_vertdiff ?    
     !JGJ 2013/05/31  merged COBALT into siena_201303
-    if(do_generic_age .or. do_generic_argon .or. do_generic_CFC .or. do_generic_TOPAZ .or. do_generic_ERGOM &
-       .or. do_generic_BLING .or. do_generic_miniBLING .or. do_generic_COBALT .or. do_generic_blres) then
+    if(do_generic_age .or. do_generic_argon .or. do_generic_CFC &
+       .or. do_generic_BLING .or. do_generic_COBALT .or. do_generic_blres) then
```
andrew-c-ross: Also `if(do_generic_tracer) then` here? But, abiotic and SF6 are missing here. Can we actually delete generic_tracer_vertdiff_M (in a separate PR)?

generic_tracers/generic_tracer.F90
```diff
@@ -792,8 +792,8 @@ subroutine generic_tracer_coupler_set(IOB_struc, ST,SS,rho,ilb,jlb,tau, dzt, sos
     !for each tracer in the tracer_list that has been marked by the user routine above
     !JGJ 2013/05/31  merged COBALT into siena_201303
     !
-    if(do_generic_abiotic .or. do_generic_age .or. do_generic_argon .or. do_generic_CFC .or. do_generic_SF6 .or. do_generic_TOPAZ &
-      .or. do_generic_ERGOM .or. do_generic_BLING .or. do_generic_miniBLING .or. do_generic_COBALT .or. do_generic_blres) &
+    if(do_generic_abiotic .or. do_generic_age .or. do_generic_argon .or. do_generic_CFC .or. do_generic_SF6  &
+      .or. do_generic_BLING .or. do_generic_COBALT .or. do_generic_blres) &
```
andrew-c-ross: And here

generic_tracers/generic_tracer.F90
andrew-c-ross: #48

yichengt900: @andrew-c-ross, good point. Personally, I would love to remove unused modules or subroutines like `generic_tracer_vertdiff_M`, but I would like to hear from other people as well. Perhaps we can discuss this this afternoon.

yichengt900:
Hi @andrew-c-ross, I have addressed most of your comments except for the one regarding `generic_tracer_vertdiff_M` (which I will handle in a separate PR). Feel free to let me know if you have any further thoughts.

andrew-c-ross (approved):

Issue #38, opened 2024-04-06
3D arrays used for 2D bottom flux and sediment storage "tracers"?


charliestock:
I believe that generic tracers is currently passing bottom flux information as a 3D tracers but these are 2D variables.  The potential expense of this has been compounded by the addition of many new variables with boundary fluxes:

https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/e499f4caa67d9ce358b08c3ff9d87e6ba521352d/generic_tracers/generic_COBALT.F90#L7254-L7418

Niki, am I correct on this?  If so, could we make the code more memory efficient by allowing for 2D tracers?  How difficult do you think this would be?

Thanks,
Charlie

yichengt900:
#46 is merged. Close this issue now.

yichengt900:
Based on today's discussion, I will revert the `bottom flux 3d to 2d change` for now. We will reconsider once we have a better idea of its impact on the restart file.

yichengt900:
After further investigation, this may require some infrastructure changes. Since the memory impact of converting sediment from 3D to 2D is not that significant in the current COBALT, we will close this issue for now.

PR #46, opened 2024-04-12
Change COBALT bottom fluxes from 3d array to 2d

yichengt900:
This PR resolves issue #38 by converting the COBALT bottom fluxes array from 3D to 2D, enhancing memory efficiency.
this PR is solely for code refactoring and does not alter the modeled results.


andrew-c-ross (approved):
At first I was surprised that this works, since the vertdiff routine uses %btf, but then I remembered vertdiff only gets applied to prognostic tracers and these are all diagnostic tracers, so this checks out :+1: 

The only side effect I can see is that enabling the bottom flux automatically creates potentially confusing diagnostics with names like "lithdet_btf_btf".

yichengt900:
@andrew-c-ross, thank you. Regarding the side effects, perhaps we could consider renaming those diagnostics? For example, renaming `cadet_arag_btf` to` b_cadet_arag` to avoid confusion. What are your thoughts?

andrew-c-ross:
> Regarding the side effects, perhaps we could consider renaming those diagnostics? For example, renaming `cadet_arag_btf` to` b_cadet_arag` to avoid confusion. 

I think it's fine to leave everything as is. The *_btf_btf diagnostic only appears if you look for it in the available_diags file.

andrew-c-ross:
I looked at one of my older runs, for other variables that have a bottom flux, I don't see that bottom flux saved in the restarts (for example, I don't see no3_btf in the restarts). This could confirm Charlie's suspicion that the new diagnostic _btfs aren't being saved in the restarts.

PR #57, opened 2024-05-14
add parameters for case 2 (coastal) waters

charliestock:
During development of the NWA12, we added a salinity threshold and a depth threshold for defining "case 2" coastal waters.  If these conditions were met, the opacity of the water was augmented from the value determined by the chlorophyll based Manizza scheme so that the transmission profile was more consistent with values from coastal waters that have high concentrations of suspended matter and colored dissolved organic material (CDOM).  In NWA12, this enhanced attenuation was primarily added to avoid over-estimating primary production in turbid nearshore waters near the Mississippi River.  This could be particularly problematic near the bottom, since primary production creates oxygen, potentially leading to underestimation of the Gulf of Mexico hypoxic region.

In the NWA12 described in Ross et al. (2023), the depth threshold was set to 30m, the salinity threshold was set to 30 PSU and the opacity was augmented by 0.05 m-1.  These values, however, were "hard coded".  This pull request would establish parameters for both thresholds (case2_depth, case2_salt) and the added opacity (case2_opac_add), allowing values to be set more transparently.  I have set these to the same values that were hard coded in NWA12, so this should not change answers, but I think there are two issues to consider moving forward:

1) What should the default values of these parameters be?  I lean toward setting the values to "0" by default, and allowing for region-specific settings as needed.  It would, however, be good to get input from the group.

2) While this implementation allows for the effects of coastal waters to be felt by the BGC, it does not impact the penetration of heat in MOM6.  A more satisfying long-term solution would be implementing a radiative transfer scheme that accounts for both coastal and open ocean environments, and does so consistently across the physical and BGC components.   I can open an issue on this latter point.

### DIFF
--- a/generic_tracers/cobalt_types.F90
+++ b/generic_tracers/cobalt_types.F90
@@ -429,6 +429,7 @@ module cobalt_types
 
      logical  ::       &
           init,             &                  ! If tracers should be initializated
+          do_case2_mod,     &                  ! Flag to include augmented opacity in shallow/river-influenced systems
           force_update_fluxes,&                ! If OCMIP2 tracers fluxes should be updated every coupling timesteps
                                                !    when update_from_source is not called every coupling timesteps
                                                !    as is the case with MOM6  THERMO_SPANS_COUPLING option
--- a/generic_tracers/cobalt_types.F90
+++ b/generic_tracers/cobalt_types.F90
@@ -461,6 +461,9 @@ module cobalt_types
           zmld_ref,         &
           densdiff_mld,     &
           irrad_day_thresh, &
+          case2_depth,      & ! depth threshold for case 2 (coastal) waters
+          case2_salt,       & ! salt threshold for case 2 (coastal) waters
+          case2_opac_add,   & ! added opacity for case 2 (coastal) waters
           min_daylength,    &
           gamma_irr_mem_dp, &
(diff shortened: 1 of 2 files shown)


yichengt900:
@charliestock , thanks for the detailed description and the changes look good to me. We can use today's meeting to discuss the second point you mentioned.

charliestock:
Hi folks,

I have added the case2 as a logical as Andrew suggested.  I am still running some tests, but all seems to have gone well so far.  

@yichengt900  I wasn't clear on whether the second commit from this pull request has passed your testing.  Is it OK?
@andrew-c-ross I believe I have implemented the parameter options as you suggested, but please let me know if I hit the mark.

yichengt900:
Hi @charliestock , Your second commit passed the test. I believe @andrew-c-ross left a comment for your second commit, which can be found [here](https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/commit/f41670de83d0455c04938c0f2cde6a0756823b52#r142048090).

charliestock:
Got it.  Now I understand. I have implemented in the way Andrew suggested to avoid a field table overwrite in cases where do_case2_mod is false.   Good suggestion Andrew.

yichengt900 (approved):
Looks good to me, approve.

Issue #55, opened 2024-04-30
carbonite -> carbonate?


andrew-c-ross:
I think the diagnostic description for cased should be calcium carbon**a**te, not calcium carbon**i**te. 

https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/76576059ee6126b8751f475da62fa44543aad937/generic_tracers/cobalt_reg_diag.F90#L1919

PR #58, opened 2024-05-14
cleanup phytoplankton uptake section

charliestock:
This pull request just cleans up the uptake code and adds a few comments.  I have also suggested removing the diagnostic "net_phyto_resp", which we don't really use for anything and can be easily calculated from the growth rate.   We should check to ensure this is not in our standard diagnostic tables.

These changes should not change answers.

### DIFF
--- a/generic_tracers/cobalt_reg_diag.F90
+++ b/generic_tracers/cobalt_reg_diag.F90
@@ -612,10 +612,6 @@ module COBALT_reg_diag
     phyto(LARGE)%id_jprod_n = register_diag_field(package_name, vardesc_temp%name, axes(1:3),&
          init_time, vardesc_temp%longname,vardesc_temp%units, missing_value = missing_value1)
 
-    vardesc_temp = vardesc("net_phyto_resp","Net phytoplankton respiration layer integral",'h','L','s','mol m-2 s-1','f')
-    cobalt%id_net_phyto_resp = register_diag_field(package_name, vardesc_temp%name, axes(1:3),&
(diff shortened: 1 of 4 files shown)


yichengt900:
@charliestock , by checking our current [NWA12 diag_table](https://github.com/NOAA-GFDL/CEFI-regional-MOM6/blob/1d799141b6842b04d11d38abbcfde129894c7ef2/xmls/NWA12/diag_tables/diag_table_COBALT_2023-04#L290), we do output water column integrated net_phyto_resp.

yichengt900 (commented):
If we want to remove `net_phyto_resp`. We will also have to remove the following diagnostic lines in `cobalt_send_diag.F90`:

https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/96914377095b78dc271fd908ca5bddc34ebfcb04/generic_tracers/cobalt_send_diag.F90#L331-L333

https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/96914377095b78dc271fd908ca5bddc34ebfcb04/generic_tracers/cobalt_send_diag.F90#L1175-L1176

charliestock:
I think net_phyto_resp and wc_net_phyto_resp have now been successfully removed from the code.  I have also confirmed that it is indeed redundant with other diagnostics.  The only thing that we'll need to ensure is that the diagnostic "wc_net_phyto_resp" is removed from all of our diagnostic tables.

yichengt900 (approved):
Changes looks good, @charliestock , I also used this change to fix #55. Approve.

PR #59, opened 2024-05-14
Add support for a COBALT_param_doc file

theresa-cordero:
Using the file parser from MOM6, add the option to read parameters from COBALT_input and COBALT_override files and write parameters to COBALT_param_doc files. 

With these changes, COBALT_input and COBALT_override files must be listed in the input namelist. As a result all existing experiments will need their namelists updated and COBALT_input files generated. The files can be blank and the default values will be used.  
 ```
&cobalt_input_nml
             parameter_filename = 'INPUT/COBALT_input','INPUT/COBALT_override'
/
```

Descriptions of the parameters have not been added with this PR and some parameters lack units. For the description, the parameter name has been used but should be updated to include a note about what the parameter is used for and a recommended range (or that the parameter not be adjusted). 

As @amoebaliz suggested, here are the needed changes to make to an xml to use the parameter doc file.

1.  In the `<input>` section for each experiment or for the primary experiment, add the new namelist option:
```
<namelist name = cobalt_input_nml >
  parameter_filename = 'INPUT/COBALT_input','INPUT/COBALT_override'
</namelist>
```
2. You can create a `COBALT_input` file in the same location as your `MOM_input` file and link to is in a similar way, for example:
```
<dataSource site="ncrc">$(ESMG_H)/ESMG-configs/Arctic6/mid_12/MOM_input</dataSource>
<dataSource site="ncrc">$(ESMG_H)/ESMG-configs/Arctic6/mid_12/COBALT_input</dataSource>
```
3.  Alternatively you can write a `COBALT_input` file directly from the xml. Toward the end of the `input` section, where the `MOM_override` file is defined in you can add:
```
<csh type='always'>
<![CDATA[
truncate -s 0 $work/INPUT/MOM_override
cat > $work/INPUT/MOM_override << MOM_OVERRIDE_EOF
MOM_OVERRIDE_EOF

truncate -s 0 $work/INPUT/COBALT_input
cat > $work/INPUT/COBALT_input << COBALT_INPUT_EOF
! Put the contents of your COBALT_input file here
COBALT_INPUT_EOF
]]>
</csh>
```
This is how and where you can add a COBALT_override file
```
truncate -s 0 $work/INPUT/COBALT_override
cat > $work/INPUT/COBALT_override << COBALT_OVERRIDE_EOF
! Put the contents of your COBALT_override file here.
! For parameters which are not specified in the COBALT_input file, just set the new value, ex:
! do_case2_mod = True
! For parameters which are specified in the COBALT_input file, specify an override, ex:
! #override do_case2_mod = True
COBALT_OVERRIDE_EOF
```

5. Finally, you can update the post processing section to include the COBALT_parameter_doc files.
```
    <postProcess>
      <csh><![CDATA[
         cd $work
         #Make a directory to trick FRE to pick up and archive in ascii
         mkdir extra.results
         mv *velocity_truncations MOM_parameter_doc* SIS_parameter_doc* COBALT_parameter_doc* seaice.stats* ocean.stats* extra.results/
         cp diag_table extra.results/.
         cp $scriptName extra.results/runscript.csh
         #When the ocean uses a mask_table the ocean_static.nc file produced by the model run has holes in coordinates (geolon,geolat)
         #This causes problems for ferret and python tools for analysis.
         #Copy a non-masked version of ocean_static.nc to be saved as a history file to be used by the analysis scipts .
         #cp $work/INPUT/ocean_static_no_mask_table.nc $work/
      ]]></csh>
      <xi:include xpointer="xpointer(//freInclude[@name='MOM6_postprocess']/postProcess/node())"/>
    </postProcess>
```

### DIFF
--- a/generic_tracers/cobalt_param_doc.F90
+++ b/generic_tracers/cobalt_param_doc.F90
@@ -1,0 +1,86 @@
+! this module will contain a few main routines: one to get the
+module cobalt_param_doc 
+
+use MOM_cpu_clock,     only : cpu_clock_id, cpu_clock_begin, cpu_clock_end, CLOCK_COMPONENT
+use MOM_data_override, only : data_override
+use MOM_domains,       only : domain2D, same_domain
+use MOM_error_handler, only : MOM_error, FATAL, WARNING, callTree_enter, callTree_leave
+use MOM_file_parser,   only : param_file_type, open_param_file, close_param_file
+use MOM_file_parser,   only : read_param, get_param, log_param, log_version
+use MOM_io,            only : file_exists, close_file, slasher, ensembler
+use MOM_io,            only : open_namelist_file, check_nml_error
+use MOM_time_manager,  only : time_type, time_type_to_real, real_to_time_type
+use MOM_time_manager,  only : operator(+), operator(-), operator(>)
+use posix, only : mkdir, stat, stat_buf
+
+implicit none ; private
+
+public get_COBALT_param_file
+
+!> Container for paths and parameter file names.
+!type, public :: directories
+!  character(len=240) :: &
+!    restart_input_dir = ' ',& !< The directory to read restart and input files.
+!    restart_output_dir = ' ',&!< The directory into which to write restart files.
+!    output_directory = ' '    !< The directory to use to write the model output.
+!  character(len=2048) :: &
+!    input_filename  = ' '     !< A string that indicates the input files or how
+!                              !! the run segment should be started.
+!end type directories
+
+contains
+!! This subroutine is analagous to the get_mom_input routine within the MOM_get_input module. There are additional
+!> arguments used in the MOM version which are commented out and not used here. These features may be needed as we 
+!> add more complexity to the parameter files in COBALT so I have commented them out rather than deleting them.
+subroutine get_COBALT_param_file(param_file)
+  type(param_file_type), optional, intent(out) :: param_file   !< A structure to parse for run-time parameters.
+  !type(directories),     optional, intent(out) :: dirs         !< Container for paths and parameter filenames.
+  !logical,               optional, intent(in)  :: check_params !< If present and False will stop error checking for
+  !                                                             !! run-time parameters.
+  !character(len=*),      optional, intent(in)  :: default_input_filename !< If present, is the value assumed for
+  !                                                             !! input_filename if input_filename is no listed
+  !                                                             !! in the namelist MOM_input_nml.
+  ! Local variables
+  integer, parameter :: npf = 5 ! Maximum number of parameter files
+
+  character(len=240) :: &
+    output_directory = ' ', &      ! Directory to use to write the model output.
+    parameter_filename(npf) = ' '  ! List of files containing parameters.
+
+  character(len=2048) :: &
+    input_filename             ! A string that indicates the input files or how
+                               ! the run segment should be started.
+  character(len=240) :: output_dir
+  integer :: unit, io, ierr, valid_param_files
+
+  type(stat_buf) :: buf
+
+  namelist /cobalt_input_nml/ parameter_filename
+
+  ! Open namelist
+  if (file_exists('input.nml')) then
+    unit = open_namelist_file(file='input.nml')
+  else
+    call MOM_error(FATAL,'Required namelist file input.nml does not exist.')
+  endif
+
+  ! Read namelist parameters
+  ierr=1 ; do while (ierr /= 0)
+    read(unit, nml=cobalt_input_nml, iostat=io, end=10)
+    ierr = check_nml_error(io, 'cobalt_input_nml')
+  enddo
+10 call close_file(unit)
+
+  output_dir = trim(slasher(ensembler(output_directory)))
+  valid_param_files = 0
+  do io=1,npf
+     if (len_trim(trim(parameter_filename(io))) > 0) then
+       call open_param_file(trim(parameter_filename(io)), param_file, &
+                            component="COBALT", doc_file_dir=output_dir)
+        valid_param_files = valid_param_files + 1
+     endif
+   enddo
+
+end subroutine get_COBALT_param_file
+
+end module cobalt_param_doc 
--- a/generic_tracers/cobalt_param_doc.F90
+++ b/generic_tracers/cobalt_param_doc.F90
@@ -78,7 +78,7 @@ subroutine get_COBALT_param_file(param_file)
        call open_param_file(trim(parameter_filename(io)), param_file, &
                             component="COBALT", doc_file_dir=output_dir)
         valid_param_files = valid_param_files + 1
-     endif
+     endif 
    enddo
 
 end subroutine get_COBALT_param_file
--- a/generic_tracers/cobalt_param_doc.F90
+++ b/generic_tracers/cobalt_param_doc.F90
@@ -78,7 +78,7 @@ subroutine get_COBALT_param_file(param_file)
        call open_param_file(trim(parameter_filename(io)), param_file, &
                             component="COBALT", doc_file_dir=output_dir)
         valid_param_files = valid_param_files + 1
-     endif 
+     endif
    enddo
 
 end subroutine get_COBALT_param_file
--- a/generic_tracers/cobalt_types.F90
+++ b/generic_tracers/cobalt_types.F90
@@ -55,10 +55,16 @@ module cobalt_types
   integer, parameter, public :: SMALL      = 4 !< ID for small phytoplankton 
 
   real, parameter, public :: sperd = 24.0 * 3600.0    !< number of seconds in a day (sec)
+  real, parameter, public :: I_sperd = 1.0/sperd      !< inverse of number of seconds in a day (sec)
   real, parameter, public :: spery = 365.25 * sperd   !< number of seconds in a year (sec)
+  real, parameter, public :: I_spery = 1.0/spery      !< inverse of number of seconds in a year (sec)
   real, parameter, public :: epsln=1.0e-30            !< small, but non-zero value for numerical stability
   real, parameter, public :: missing_value1=-1.0e+10  !< large negative value to represent missing value in diags
   real, parameter, public :: vb_nh3 = 25.             !< Liquid molar volume at boiling point for NH3 (cm3 mol−1)
+  real, parameter, public :: micromol2mol = 1.0e6     !< convert micromoles to moles (used in concentration parameters)
+  real, parameter, public :: c2n = 106.0/16.0         !< convert nutrient ratios relative to carbon to ratios relative to nitrogen
+  real, parameter, public :: micromolQpersec2W = 2.77e18/6.022e17 !< convert photosynthetically available radiation (micromole quanta
+                                                      !! per-second) to watts. 
 
   !> An auxiliary type for storing varible names
   type vardesc
--- a/generic_tracers/generic_COBALT.F90
+++ b/generic_tracers/generic_COBALT.F90
@@ -564,14 +564,22 @@ contains
     call get_param(param_file, "generic_COBALT", "alpha_Md_ll",         phyto(MEDIUM)%alpha_ll,         "alpha_Md_ll",         units="g C g chl-1 s-1", default= 1.6e-5  * 2.77e18 / 6.022e17)! g C g Chl-1 sec-1 (W m-2)-1
     call get_param(param_file, "generic_COBALT", "alpha_Sm_ll",         phyto(SMALL)%alpha_ll,          "alpha_Sm_ll",         units="g C g chl-1 s-1", default= 3.2e-5  * 2.77e18 / 6.022e17)! g C g Chl-1 sec-1 (W m-2)-1
     call get_param(param_file, "generic_COBALT", "kappa_eppley",        cobalt%kappa_eppley,            "kappa_eppley",        units="deg C-1",         default= 0.063)                   ! deg C-1
-    call get_param(param_file, "generic_COBALT", "P_C_max_Di_hl",       phyto(DIAZO)%P_C_max_hl,        "P_C_max_Di_hl",       units="s-1",             default= 0.6/sperd)          ! s-1
-    call get_param(param_file, "generic_COBALT", "P_C_max_Lg_hl",       phyto(LARGE)%P_C_max_hl,        "P_C_max_Lg_hl",       units="s-1",             default= 1.0/sperd)          ! s-1
-    call get_param(param_file, "generic_COBALT", "P_C_max_Md_hl",       phyto(MEDIUM)%P_C_max_hl,       "P_C_max_Md_hl",       units="s-1",             default= 1.1/sperd)         ! s-1
-    call get_param(param_file, "generic_COBALT", "P_C_max_Sm_hl",       phyto(SMALL)%P_C_max_hl,        "P_C_max_Sm_hl",       units="s-1",             default= 1.0/sperd)          ! s-1
-    call get_param(param_file, "generic_COBALT", "P_C_max_Di_ll",       phyto(DIAZO)%P_C_max_ll,        "P_C_max_Di_ll",       units="s-1",             default= 0.3/sperd)     ! s-1
-    call get_param(param_file, "generic_COBALT", "P_C_max_Lg_ll",       phyto(LARGE)%P_C_max_ll,        "P_C_max_Lg_ll",       units="s-1",             default= 0.5/sperd)     ! s-1
-    call get_param(param_file, "generic_COBALT", "P_C_max_Md_ll",       phyto(MEDIUM)%P_C_max_ll,       "P_C_max_Md_ll",       units="s-1",             default= 0.55/sperd)    ! s-1
-    call get_param(param_file, "generic_COBALT", "P_C_max_Sm_ll",       phyto(SMALL)%P_C_max_ll,        "P_C_max_Sm_ll",       units="s-1",             default= 0.5/sperd)     ! s-1
+    call get_param(param_file, "generic_COBALT", "P_C_max_Di_hl",       phyto(DIAZO)%P_C_max_hl,        "P_C_max_Di_hl",       units="day-1",           default= 0.6)
+      phyto(DIAZO)%P_C_max_hl = phyto(DIAZO)%P_C_max_hl/sperd   ! s-1
+    call get_param(param_file, "generic_COBALT", "P_C_max_Lg_hl",       phyto(LARGE)%P_C_max_hl,        "P_C_max_Lg_hl",       units="day-1",           default= 1.0)
+      phyto(LARGE)%P_C_max_hl = phyto(LARGE)%P_C_max_hl/sperd   ! s-1
+    call get_param(param_file, "generic_COBALT", "P_C_max_Md_hl",       phyto(MEDIUM)%P_C_max_hl,       "P_C_max_Md_hl",       units="day-1",           default= 1.1)
+      phyto(MEDIUM)%P_C_max_hl = phyto(MEDIUM)%P_C_max_hl/sperd   ! s-1
+    call get_param(param_file, "generic_COBALT", "P_C_max_Sm_hl",       phyto(SMALL)%P_C_max_hl,        "P_C_max_Sm_hl",       units="day-1",           default= 1.0)
+      phyto(SMALL)%P_C_max_hl = phyto(SMALL)%P_C_max_hl/sperd   ! s-1
+    call get_param(param_file, "generic_COBALT", "P_C_max_Di_ll",       phyto(DIAZO)%P_C_max_ll,        "P_C_max_Di_ll",       units="day-1",           default= 0.3)
+      phyto(DIAZO)%P_C_max_ll = phyto(DIAZO)%P_C_max_ll/sperd   ! s-1
+    call get_param(param_file, "generic_COBALT", "P_C_max_Lg_ll",       phyto(LARGE)%P_C_max_ll,        "P_C_max_Lg_ll",       units="day-1",           default= 0.5)
+      phyto(LARGE)%P_C_max_ll = phyto(LARGE)%P_C_max_ll/sperd   ! s-1
+    call get_param(param_file, "generic_COBALT", "P_C_max_Md_ll",       phyto(MEDIUM)%P_C_max_ll,       "P_C_max_Md_ll",       units="day-1",           default= 0.55)
+      phyto(MEDIUM)%P_C_max_ll = phyto(MEDIUM)%P_C_max_ll/sperd   ! s-1
+    call get_param(param_file, "generic_COBALT", "P_C_max_Sm_ll",       phyto(SMALL)%P_C_max_ll,        "P_C_max_Sm_ll",       units="day-1",           default= 0.5)
+      phyto(SMALL)%P_C_max_ll = phyto(SMALL)%P_C_max_ll/sperd   ! s-1
     call get_param(param_file, "generic_COBALT", "numlightadapt",       cobalt%numlightadapt,           "numlightadapt",       units="",                default= 10)               ! dimensionless
     call get_param(param_file, "generic_COBALT", "thetamax_Di",         phyto(DIAZO)%thetamax,          "thetamax_Di",         units="g chl g C-1",     default= 0.035)                  ! g Chl g C-1
     call get_param(param_file, "generic_COBALT", "thetamax_Lg",         phyto(LARGE)%thetamax,          "thetamax_Lg",         units="g chl g C-1",     default= 0.07)                  ! g Chl g C-1
@@ -585,16 +593,23 @@ contains
     call get_param(param_file, "generic_COBALT", "bresp_frac_strat_Lg", phyto(LARGE)%bresp_frac_strat,  "bresp_frac_strat_Lg", units="",                default= 0.01)   ! none
     call get_param(param_file, "generic_COBALT", "bresp_frac_strat_Md", phyto(MEDIUM)%bresp_frac_strat, "bresp_frac_strat_Md", units="",                default= 0.01)  ! none
     call get_param(param_file, "generic_COBALT", "bresp_frac_strat_Sm", phyto(SMALL)%bresp_frac_strat,  "bresp_frac_strat_Sm", units="",                default= 0.01)   ! none
-    call get_param(param_file, "generic_COBALT", "sink_max_Di",         phyto(DIAZO)%sink_max,          "sink_max_Di",         units="m sec-1",     default= 1.0/sperd)              ! m sec-1
-    call get_param(param_file, "generic_COBALT", "sink_max_Lg",         phyto(LARGE)%sink_max,          "sink_max_Lg",         units="m sec-1",     default= 5.0/sperd)              ! m sec-1
-    call get_param(param_file, "generic_COBALT", "sink_max_Md",         phyto(MEDIUM)%sink_max,         "sink_max_Md",         units="m sec-1",     default= 1.0/sperd)             ! m sec-1
-    call get_param(param_file, "generic_COBALT", "sink_max_Sm",         phyto(SMALL)%sink_max,          "sink_max_Sm",         units="m sec-1",     default= 0.0/sperd)              ! m sec-1
+    call get_param(param_file, "generic_COBALT", "sink_max_Di",         phyto(DIAZO)%sink_max,          "sink_max_Di",         units="m day-1",     default= 1.0)
+      phyto(DIAZO)%sink_max = phyto(DIAZO)%sink_max/sperd ! m sec-1
+    call get_param(param_file, "generic_COBALT", "sink_max_Lg",         phyto(LARGE)%sink_max,          "sink_max_Lg",         units="m day-1",     default= 5.0)
+      phyto(LARGE)%sink_max = phyto(LARGE)%sink_max/sperd ! m sec-1
+    call get_param(param_file, "generic_COBALT", "sink_max_Md",         phyto(MEDIUM)%sink_max,         "sink_max_Md",         units="m day-1",     default= 1.0)
+      phyto(MEDIUM)%sink_max = phyto(MEDIUM)%sink_max/sperd ! m sec-1
+    call get_param(param_file, "generic_COBALT", "sink_max_Sm",         phyto(SMALL)%sink_max,          "sink_max_Sm",         units="m day-1",     default= 0.0)
+      phyto(SMALL)%sink_max = phyto(SMALL)%sink_max/sperd ! m sec-1
     call get_param(param_file, "generic_COBALT", "thetamin",            cobalt%thetamin,                "thetamin",            units="g chl g C-1", default= 0.002)                          ! g Chl g C-1
     call get_param(param_file, "generic_COBALT", "zeta",                cobalt%zeta,                    "zeta",                units="",            default= 0.05)                                   ! dimensionless
     call get_param(param_file, "generic_COBALT", "par_adj",             cobalt%par_adj,                 "par_adj",             units="",            default= 0.83)                              ! dimensionless
-    call get_param(param_file, "generic_COBALT", "gamma_irr_aclm",      cobalt%gamma_irr_aclm,          "gamma_irr_aclm",      units="s-1",         default= 1.0 / sperd)        ! s-1
-    call get_param(param_file, "generic_COBALT", "gamma_irr_mem_dp",    cobalt%gamma_irr_mem_dp,        "gamma_irr_mem_dp",    units="s-1",         default= 0.1/sperd)        ! s-1
-    call get_param(param_file, "generic_COBALT", "gamma_mu_mem",        cobalt%gamma_mu_mem,            "gamma_mu_mem",        units="s-1",         default= 1.0 / sperd)            ! s-1
+    call get_param(param_file, "generic_COBALT", "gamma_irr_aclm",      cobalt%gamma_irr_aclm,          "gamma_irr_aclm",      units="day-1",       default= 1.0)
+      cobalt%gamma_irr_aclm = cobalt%gamma_irr_aclm/sperd ! s-1
+    call get_param(param_file, "generic_COBALT", "gamma_irr_mem_dp",    cobalt%gamma_irr_mem_dp,        "gamma_irr_mem_dp",    units="day-1",       default= 0.1)
+      cobalt%gamma_irr_mem_dp= cobalt%gamma_irr_mem_dp/sperd ! s-1
+    call get_param(param_file, "generic_COBALT", "gamma_mu_mem",        cobalt%gamma_mu_mem,            "gamma_mu_mem",        units="day-1",       default= 1.0)
+      cobalt%gamma_mu_mem= cobalt%gamma_mu_mem/sperd ! s-1
     call get_param(param_file, "generic_COBALT", "ml_aclm_efold",       cobalt%ml_aclm_efold,           "ml_aclm_efold",       units="",            default= 2.5)                  ! dimensionless
     call get_param(param_file, "generic_COBALT", "zmld_ref",            cobalt%zmld_ref,                "zmld_ref",            units="m",           default= 10.0)                           ! m
     call get_param(param_file, "generic_COBALT", "densdiff_mld",        cobalt%densdiff_mld,            "densdiff_mld",        units="kg m-3",      default= 0.03)                   ! kg m-3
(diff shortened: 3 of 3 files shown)

### CODE REVIEW

**Comment by yichengt900** [comment]
Thanks @theresa-morrison . I have updated the [github actions testing workflow](https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/pull/62) to accommodate this PR. Please update your feature branch, and it should be able to pass the RT.

I can confirm that this new functionality works properly. For example, if I set `remin_ramp_scale = 100.0` in COBALT_input and use `COBALT_override` to revert it to the default value, the model results are identical to the reference.

@charliestock and anyone else, it would be great if you could double-check the changes in `generic_tracers/generic_COBALT.F90` and ensure you are comfortable with these updates.

**Comment by theresa-cordero** [comment]
As @amoebaliz suggested, here are the needed changes to make to an xml to use the parameter doc file.

1.  In the `<input>` section for each experiment or for the primary experiment, add the new namelist option:
```
<namelist name = cobalt_input_nml >
  parameter_filename = 'INPUT/COBALT_input','INPUT/COBALT_override'
</namelist>
```
2. You can create a `COBALT_input` file in the same location as your `MOM_input` file and link to is in a similar way, for example:
```
<dataSource site="ncrc">$(ESMG_H)/ESMG-configs/Arctic6/mid_12/MOM_input</dataSource>
<dataSource site="ncrc">$(ESMG_H)/ESMG-configs/Arctic6/mid_12/COBALT_input</dataSource>
```
3.  Alternatively you can write a `COBALT_input` file directly from the xml. Toward the end of the `input` section, where the `MOM_override` file is defined in you can add:
```
<csh type='always'>
<![CDATA[
truncate -s 0 $work/INPUT/MOM_override
cat > $work/INPUT/MOM_override << MOM_OVERRIDE_EOF
MOM_OVERRIDE_EOF

truncate -s 0 $work/INPUT/COBALT_input
cat > $work/INPUT/COBALT_input << COBALT_INPUT_EOF
! Put the contents of your COBALT_input file here
COBALT_INPUT_EOF
]]>
</csh>
```
This is how and where you can add a COBALT_override file
```
truncate -s 0 $work/INPUT/COBALT_override
cat > $work/INPUT/COBALT_override << COBALT_OVERRIDE_EOF
! Put the contents of your COBALT_override file here.
! For parameters which are not specified in the COBALT_input file, just set the new value, ex:
! do_case2_mod = True
! For parameters which are specified in the COBALT_input file, specify an override, ex:
! #override do_case2_mod = True
COBALT_OVERRIDE_EOF
```

5. Finally, you can update the post processing section to include the COBALT_parameter_doc files.
```
    <postProcess>
      <csh><![CDATA[
         cd $work
         #Make a directory to trick FRE to pick up and archive in ascii
         mkdir extra.results
         mv *velocity_truncations MOM_parameter_doc* SIS_parameter_doc* COBALT_parameter_doc* seaice.stats* ocean.stats* extra.results/
         cp diag_table extra.results/.
         cp $scriptName extra.results/runscript.csh
         #When the ocean uses a mask_table the ocean_static.nc file produced by the model run has holes in coordinates (geolon,geolat)
         #This causes problems for ferret and python tools for analysis.
         #Copy a non-masked version of ocean_static.nc to be saved as a history file to be used by the analysis scipts .
         #cp $work/INPUT/ocean_static_no_mask_table.nc $work/
      ]]></csh>
      <xi:include xpointer="xpointer(//freInclude[@name='MOM6_postprocess']/postProcess/node())"/>
    </postProcess>
```

**Comment by theresa-cordero** [comment]
With the latest commit, parameters with default values that included a conversion from per-day or per-year ito per-second in the default have been updated so that there default is the per-day or per-year value. 

These value are better understood in the community which will make it easier for users to adjust them. 

In addition, avoiding division in calculation of the default is important for producing accurate values in the parameter doc files. Otherwise, when the rounded values from the parameter doc file are specified rather than the default the difference changes the model results.

This commit did not impact the parameters with calculations in the default that appeared to be related to stoichiometry. These parameters should be clearly labeled in their descriptions possibly with a warning that users should not adjust these values.

File: generic_tracers/generic_COBALT.F90 [hunk 1]
```diff
@@ -496,456 +508,507 @@ subroutine user_add_params
...
-    call g_tracer_add_param('k_no3_Md', phyto(MEDIUM)%k_no3, 1.0e-6)                 ! mol NO3 kg-1
-    call g_tracer_add_param('k_no3_Sm', phyto(SMALL)%k_no3,  5.0e-7)                  ! mol NO3 kg-1
-    call g_tracer_add_param('k_no3_Di', phyto(DIAZO)%k_no3,  5.0e-6)                  ! mol NO3 kg-1
-    call g_tracer_add_param('k_po4_Di', phyto(DIAZO)%k_po4,  1.0e-7)                  ! mol PO4 kg-1
-    call g_tracer_add_param('k_po4_Lg', phyto(LARGE)%k_po4,  5.0e-8)                  ! mol PO4 kg-1
-    call g_tracer_add_param('k_po4_Md', phyto(MEDIUM)%k_po4, 2.0e-8)                  ! mol PO4 kg-1
-    call g_tracer_add_param('k_po4_Sm', phyto(SMALL)%k_po4,  1.0e-8)                  ! mol PO4 kg-1
-    call g_tracer_add_param('k_sio4_Lg',phyto(LARGE)%k_sio4, 2.0e-6)                  ! mol SiO4 kg-1
-    call g_tracer_add_param('k_sio4_Md',phyto(MEDIUM)%k_sio4,1.0e-6)                  ! mol SiO4 kg-1
-    call g_tracer_add_param('k_fe_2_n_Di', phyto(DIAZO)%k_fe_2_n, 12.0e-6 * 106.0 / 16.0)   ! mol Fe mol N-1
```
theresa-cordero, line 532: @charliestock @amoebaliz @andrew-c-ross  
Are these parameter defaults that depend on stoichiometry or is there another reason for the division?

generic_tracers/generic_COBALT.F90
```diff
@@ -496,456 +508,507 @@ subroutine user_add_params
...
     !-----------------------------------------------------------------------
     ! Other stoichiometry
     !-----------------------------------------------------------------------
     !
-    call g_tracer_add_param('p_2_n_static', cobalt%p_2_n_static, .false. )
-    call g_tracer_add_param('c_2_n', cobalt%c_2_n, 106.0 / 16.0)
-    call g_tracer_add_param('alk_2_n_denit', cobalt%alk_2_n_denit, 552.0/472.0)             ! eq. alk mol NO3-1
-    call g_tracer_add_param('alk_2_nh4_amx', cobalt%alk_2_nh4_amx, 3732.0/8030.0)           ! eq. alk mol NH4-1
-    call g_tracer_add_param('p_2_n_static_Di', phyto(DIAZO)%p_2_n_static,1.0/40.0 )         ! mol P mol N-1
-    call g_tracer_add_param('p_2_n_static_Lg', phyto(LARGE)%p_2_n_static,1.0/14.0 )         ! mol P mol N-1
```
theresa-cordero, line 618: @charliestock @amoebaliz @andrew-c-ross 
Can these 1.0/X parameters be changed to X and then we can do the 1.0/X calculation on the next line? 
Some of the resulting defaults cannot be exactly represented in the parameter file.

yichengt900: @theresa-morrison, I will let the others answer your concern, but I think that we probably do not want users to realize that they can easily/directly change these `classic stoichiometry` parameters at this moment. Perhaps there is an option in the `get_param` call to do so.

theresa-cordero:
There are two relevant options for `get_param` that we can use here:
`do_not_read` : If present and true, do not read a value for this parameter, although it might be logged.
`scale` : A scaling factor that the parameter is multiplied by before it is returned.

`do_not read` would be good for the stoichiometry parameters that users may want to make sure are consistent, but should not be able to easily (or accidentally) change. 

`scale` is similar to what Xiao suggested in the documentation meeting yesterday, and would be a good option for the time unit conversions that some of the parameters have.

generic_tracers/generic_COBALT.F90
charliestock: Yes - values in the literature are generally given in units of iron to carbon.  Since nitrogen serves as the main currency in COBALT and other BGC models, we covert to iron to nitrogen by using the Redfield ratio of 106 carbons per unit of nitrogen.

This could be handled with a "sperd"-like conversion c2n = 106/16?

generic_tracers/generic_COBALT.F90
charliestock: I think we a reasonable level of accuracy is fine here (i.e., we could convert to a decimal value if needed).  That said, I agree with @yichengt900 that there are some values that should not be changed.  @theresa-morrison, my understanding from our conversation yesterday was that we could set this with an option not to read external values for certain parameters that should not be changed?

charliestock: I obviously didn't read your next comment.....please see additional thoughts below.

generic_tracers/cobalt_param_doc.F90
```diff
@@ -0,0 +1,86 @@
...
+subroutine get_COBALT_param_file(param_file)
+  type(param_file_type), optional, intent(out) :: param_file   !< A structure to parse for run-time parameters.
+  !type(directories),     optional, intent(out) :: dirs         !< Container for paths and parameter filenames.
+  !logical,               optional, intent(in)  :: check_params !< If present and False will stop error checking for
+  !                                                             !! run-time parameters.
+  !character(len=*),      optional, intent(in)  :: default_input_filename !< If present, is the value assumed for
+  !                                                             !! input_filename if input_filename is no listed
+  !                                                             !! in the namelist MOM_input_nml.
+  ! Local variables
+  integer, parameter :: npf = 5 ! Maximum number of parameter files
```
charliestock, line 44: Is there a case where we would need more than one?

generic_tracers/generic_COBALT.F90
```diff
@@ -496,456 +508,507 @@ subroutine user_add_params
...
-    call g_tracer_add_param('gamma_cadet_calc',cobalt%gamma_cadet_calc,cobalt%wsink/1343.0)  ! s-1
-    call g_tracer_add_param('kappa_sidet',  cobalt%kappa_sidet, 0.063 )                      ! deg C -1
-    call g_tracer_add_param('gamma_sidet',  cobalt%gamma_sidet, cobalt%wsink / 1.0e4 )       ! s-1
-    call g_tracer_add_param('phi_lith' ,  cobalt%phi_lith, 0.002)                            ! dimensionless
-    call g_tracer_add_param('k_lith',  cobalt%k_lith, 0.5/spery )                            ! s-1
-    call g_tracer_add_param('bottom_thickness',cobalt%bottom_thickness, 1.0 )                ! m
-    call g_tracer_add_param('z_sed',  cobalt%z_sed, 0.1 )                                    ! m
-    call g_tracer_add_param('k_no3_denit',cobalt%k_no3_denit,1.0e-6)                         ! mol NO3 kg-1
-    call g_tracer_add_param('z_burial',cobalt%z_burial,10.0)                                 ! m
+    call get_param(param_file, "generic_COBALT", "k_o2",             cobalt%k_o2,              "k_o2",             units="mol O2 kg-1   ", default= 8.0e-6)                                     ! mol O2 kg-1
```
charliestock: Note that some concentrations are currently entered as moles kg-1 (i.e., after division by 1e6), while others are entered in micromoles kg-1 units and then converted.  We should likely uniformly do the latter?

generic_tracers/generic_COBALT.F90
```diff
@@ -496,456 +508,507 @@ subroutine user_add_params
...
+    call get_param(param_file, "generic_COBALT", "gamma_cadet_arag", cobalt%gamma_cadet_arag,  "gamma_cadet_arag", units="s-1           ", default= cobalt%wsink/760.0)   ! s-1
+    call get_param(param_file, "generic_COBALT", "gamma_cadet_calc", cobalt%gamma_cadet_calc,  "gamma_cadet_calc", units="s-1           ", default= cobalt%wsink/1343.0)  ! s-1
+    call get_param(param_file, "generic_COBALT", "kappa_sidet",      cobalt%kappa_sidet,       "kappa_sidet",      units="deg C -1      ", default= 0.063 )                      ! deg C -1
+    call get_param(param_file, "generic_COBALT", "gamma_sidet",      cobalt%gamma_sidet,       "gamma_sidet",      units="s-1           ", default= cobalt%wsink / 1.0e4 )       ! s-1
+    call get_param(param_file, "generic_COBALT", "phi_lith" ,        cobalt%phi_lith,          "phi_lith" ,        units="dimensionless ", default= 0.002)                            ! dimensionless
+    call get_param(param_file, "generic_COBALT", "k_lith",           cobalt%k_lith,            "k_lith",           units="year-1        ", default= 0.5)                            ! year-1
```
charliestock: Should we require this to be entered in day-1 to prevent too many different time scales and conversions?  There are certainly long time scale processes that are commonly discussed in units of years, so maybe it is OK to have both.  We'd just need to be sure the user has ready access to all the required units.

generic_tracers/generic_COBALT.F90
```diff
@@ -496,456 +508,507 @@ subroutine user_add_params
...
     !
     !-----------------------------------------------------------------------
     ! Dissolved Organic Material
     !-----------------------------------------------------------------------
     !
-    call g_tracer_add_param('gamma_srdon',  cobalt%gamma_srdon, 1.0 / (10.0 * spery))          ! s-1
-    call g_tracer_add_param('gamma_srdop',  cobalt%gamma_srdop, 1.0 / (4.0 * spery))           ! s-1
-    call g_tracer_add_param('gamma_sldon',  cobalt%gamma_sldon, 1.0 / (90 * sperd))           ! s-1
-    call g_tracer_add_param('gamma_sldop',  cobalt%gamma_sldop, 1.0 / (90 * sperd))           ! s-1
+    call get_param(param_file, "generic_COBALT", "gamma_srdon",  cobalt%gamma_srdon, "gamma_srdon", units="1/(default*spery)", default= 10.0)
```
charliestock, line 995: The "10" here is 10 years, so the model units "sec-1".  We can eventually make this "day-1" to simplify?

generic_tracers/generic_COBALT.F90
```diff
@@ -496,456 +508,507 @@ subroutine user_add_params
...
     ! Nitrification / Anammox
     !-----------------------------------------------------------------------
     !
-    call g_tracer_add_param('gamma_nitrif',  cobalt%gamma_nitrif, gamma_nitrif / (30.0 * sperd))     ! s-1
-    call g_tracer_add_param('knh4_nitrif',  cobalt%k_nh3_nitrif, k_nh3_nitrif )                     ! moles kg-1
-    call g_tracer_add_param('irr_inhibit',  cobalt%irr_inhibit, irr_inhibit)                         ! W m-2
-    !call g_tracer_add_param('gamma_nh4amx',  cobalt%gamma_nh4amx, 0.07 / sperd)                      ! s-1
-    call g_tracer_add_param('gamma_nh4amx',  cobalt%gamma_nh4amx, 0.0 / sperd)                       ! s-1
-    call g_tracer_add_param('o2_min_amx', cobalt%o2_min_amx, 4.0e-6 )                                ! mol O2 kg-1
+    call get_param(param_file, "generic_COBALT", "gamma_nitrif",  cobalt%gamma_nitrif, "gamma_nitrif", units="gamma_nitrif /(default*sperd)", default= 30.0)     ! s-1
```
charliestock, line 1018: Need to remove Fabian's namelists references here, but we can leave this for the refinement after Theresa's foundational pull request.

charliestock:
Thanks again for the great work on this pull request Theresa.  It is an exciting development!  

The flexibility offered by "do_not_read" and "scale" sounds great.  For "scale" it seems that we have a limited number of primary conversions that may make this approach tractable:

1) Time: My inclination is to require all inputs in "days" or "days-1" and then convert to "seconds" or "seconds-1" using sperd = 86400.  There are some parameters that we have converted from "year" or "year-1" in the past, but I don't think there are many of these and it may be easiest to limit the number.  Thoughts?

2) Concentrations: We could have the input unit be micromoles kg-1 and use a uniform conversion micromol2mol = 1e6 to convert?

3) Nutrient ratios: There are a number of ratios that are commonly reported in the literature relative to carbon (i.e., iron to carbon) that are converted into ratios relative to nitrogen for the model.  A conversion c2n = 106.0/16.0 would do the trick for these.

4) Light: Most of the photosynthesis literature reports parameters, such as the initial slope of photosynthesis-irradiance curve or the compensation irradiance in "micromoles quanta" of photosynthetically available radiation.  We use a conversion from Smith and Morel (1974) to convert this to watts: under typical conditions 2.77e18 quanta sec-1 = a watt.  So there are 2.77e18/6.022e17 = 4.6 micromole quanta per sec-1 per Watt (note that 6.022e17 is a micromole).  How about something like micromolQpersec2W = 2.77e18/6.022e17?  Any suggestions on how to shorten?

I may have missed a few, but I think this would cover most of them.

Lastly, I'm wondering when the best time to accept this initial pull request is?  It seems that we'd want to accept the foundation of the pull request ASAP so that we can begin work on filling in parameter details, etc.  Do folks think we should pull things over before "scaling" settings and then work to refine and rescale together?  Or is it worth implementing/testing the rescaling first?  

Have we confirmed that this doesn't change answers for a more complete run like NWA12?

Again - excellent work Theresa and all who have contributed!

Charlie

generic_tracers/generic_COBALT.F90
```diff
@@ -496,456 +508,507 @@ subroutine user_add_params
...
+    call get_param(param_file, "generic_COBALT", "upswim_I_thresh",  zoo(1)%upswim_I_thresh,   "upswim_I_thresh",  units="", default=0.0)     ! dimensionless
+    call get_param(param_file, "generic_COBALT", "upswim_I_thresh",  zoo(2)%upswim_I_thresh,   "upswim_I_thresh",  units="", default=0.0)     ! dimensionless
+    call get_param(param_file, "generic_COBALT", "upswim_I_thresh",  zoo(3)%upswim_I_thresh,   "upswim_I_thresh",  units="", default=0.0)     ! dimensionless
+    call get_param(param_file, "generic_COBALT", "swim_max",         zoo(1)%swim_max,          "swim_max",         units="m day-1", default=100.0)
+      zoo(1)%swim_max = zoo(1)%swim_max/sperd! max swimming (m sec-1)
+    call get_param(param_file, "generic_COBALT", "swim_max",         zoo(2)%swim_max,          "swim_max",         units="m day-1", default=500.0)
+      zoo(2)%swim_max = zoo(2)%swim_max/sperd! max swimming (m sec-1)
+    call get_param(param_file, "generic_COBALT", "swim_max",         zoo(3)%swim_max,          "swim_max",         units="m day-1", default=2000.0)
+      zoo(3)%swim_max = zoo(3)%swim_max/sperd! max swimming (m sec-1)
```
yichengt900: Since we have 3 zooplankton groups here,  I suggested to change these codes as follows:"
```
    call get_param(param_file, "generic_COBALT", "upswim_chl_thresh1",zoo(1)%upswim_chl_thresh, "upswim_chl_thresh1",units="", default=0.0) ! dimensionless
    call get_param(param_file, "generic_COBALT", "upswim_chl_thresh2",zoo(2)%upswim_chl_thresh, "upswim_chl_thresh2",units="", default=0.0) ! dimensionless
    call get_param(param_file, "generic_COBALT", "upswim_chl_thresh3",zoo(3)%upswim_chl_thresh, "upswim_chl_thresh3",units="", default=0.0) ! dimensionless
    call get_param(param_file, "generic_COBALT", "upswim_I_thresh1",  zoo(1)%upswim_I_thresh,   "upswim_I_thresh1",  units="", default=0.0)     ! dimensionless
    call get_param(param_file, "generic_COBALT", "upswim_I_thresh2",  zoo(2)%upswim_I_thresh,   "upswim_I_thresh2",  units="", default=0.0)     ! dimensionless
    call get_param(param_file, "generic_COBALT", "upswim_I_thresh3",  zoo(3)%upswim_I_thresh,   "upswim_I_thresh3",  units="", default=0.0)     ! dimensionless
    call get_param(param_file, "generic_COBALT", "swim_max1",         zoo(1)%swim_max,          "swim_max1",         units="m day-1", default=100.0)
      zoo(1)%swim_max = zoo(1)%swim_max/sperd! max swimming (m sec-1)
    call get_param(param_file, "generic_COBALT", "swim_max2",         zoo(2)%swim_max,          "swim_max2",         units="m day-1", default=500.0)
      zoo(2)%swim_max = zoo(2)%swim_max/sperd! max swimming (m sec-1)
    call get_param(param_file, "generic_COBALT", "swim_max3",         zoo(3)%swim_max,          "swim_max3",         units="m day-1", default=2000.0)
      zoo(3)%swim_max = zoo(3)%swim_max/sperd! max swimming (m sec-1)
```

yichengt900:
Thanks, @theresa-morrison, for this great work, and thank you, @Charlie, for the detailed and thoughtful comments. I agree with your points and appreciate the clarity on the primary conversions.

I have a minor comment regarding the zooplankton swimming parameters that could help us maintain answers (it's not activate at this moment but the `get_param` call will lock the parameter so it's better to use different parameter names for each groups) in the NWA12 case.

Regarding the `do_case2_mod=false` setting, do we prefer this as the default?

I also support accepting this PR sooner rather than later. We can revisit the rescaling in a separate PR. This would allow other developers to benefit from the enhancements and begin working with the new functionality immediately.

Great work, @theresa-morrison, and thanks to everyone involved!

generic_tracers/generic_COBALT.F90
charliestock: Good catch YC.  We could also change up "upswim_I_thresh_smz, upswim_I_thresh_mdz, upswim_I_thresh_lgz" for even more clarity.  That said, the upward swimming isn't activated since it didn't help in the way I hoped so we may ultimately want to clean this up in favor of a more complete zooplankton vertical movement later.

theresa-cordero:
>     1. Time: My inclination is to require all inputs in "days" or "days-1" and then convert to "seconds" or "seconds-1" using sperd = 86400.  There are some parameters that we have converted from "year" or "year-1" in the past, but I don't think there are many of these and it may be easiest to limit the number.  Thoughts?

I agree with this, there are about 8 `spery` conversions. I think it would be fine to keep those around for now.
 
>     2. Concentrations: We could have the input unit be micromoles kg-1 and use a uniform conversion micromol2mol = 1e6 to convert?

I like this because it will make it very clear what the 1e6 is for. 

>     3. Nutrient ratios: There are a number of ratios that are commonly reported in the literature relative to carbon (i.e., iron to carbon) that are converted into ratios relative to nitrogen for the model.  A conversion c2n = 106.0/16.0 would do the trick for these.

I can do this, I will just search for "106.0/16.0" which should get most instances.

>     4. Light: Most of the photosynthesis literature reports parameters, such as the initial slope of photosynthesis-irradiance curve or the compensation irradiance in "micromoles quanta" of photosynthetically available radiation.  We use a conversion from Smith and Morel (1974) to convert this to watts: under typical conditions 2.77e18 quanta sec-1 = a watt.  So there are 2.77e18/6.022e17 = 4.6 micromole quanta per sec-1 per Watt (note that 6.022e17 is a micromole).  How about something like micromolQpersec2W = 2.77e18/6.022e17?  Any suggestions on how to shorten?

I think `micromolQpersec2W` is clear and I can include a reference to Smith and Morel (1974)  and some of the description here where the parameter is defined. These scaling factors will only be in the parameter read statements so it is okay if they are long. 
 
> Lastly, I'm wondering when the best time to accept this initial pull request is? It seems that we'd want to accept the foundation of the pull request ASAP so that we can begin work on filling in parameter details, etc. Do folks think we should pull things over before "scaling" settings and then work to refine and rescale together? Or is it worth implementing/testing the rescaling first?
> 
> Have we confirmed that this doesn't change answers for a more complete run like NWA12?

I can make the changes and do some basic testing on my workstation today. I have done short tests with the NWA12 but I would recommend we run ~1 year in case there are parameters that aren't active in the first month or few days of the year. I have only been looking at the ocean stats file, but we should look at the COBALT values in the standard out file as well.

I also agree with @yichengt900  that we can put the rescaling in a separate PR to expedite this. And I will change `do_case2_mod=true` back to the default.

charliestock:
OK, I like the plan of including the rescaling as a second line of code in the initial pull request, and seeing if we can take these out in a subsequent one.  We can discuss whether we want to the case2 modifications to be the default today.  I actually lean a bit toward making the default "false" since this is the case that is most consistent with the physics.

generic_tracers/cobalt_types.F90
```diff
@@ -55,10 +55,16 @@ module cobalt_types
...
+  real, parameter, public :: I_sperd = 1.0/sperd      !< inverse of number of seconds in a day (sec)
   real, parameter, public :: spery = 365.25 * sperd   !< number of seconds in a year (sec)
+  real, parameter, public :: I_spery = 1.0/spery      !< inverse of number of seconds in a year (sec)
   real, parameter, public :: epsln=1.0e-30            !< small, but non-zero value for numerical stability
   real, parameter, public :: missing_value1=-1.0e+10  !< large negative value to represent missing value in diags
   real, parameter, public :: vb_nh3 = 25.             !< Liquid molar volume at boiling point for NH3 (cm3 mol−1)
+  real, parameter, public :: micromol2mol = 1.0e6     !< convert micromoles to moles (used in concentration parameters)
+  real, parameter, public :: c2n = 106.0/16.0         !< convert nutrient ratios relative to carbon to ratios relative to nitrogen
+  real, parameter, public :: micromolQpersec2W = 2.77e18/6.022e17 !< convert photosynthetically available radiation (micromole quanta
+                                                      !! per-second) to watts.
```
theresa-cordero, line 67: @charliestock The new scale factors are added here

generic_tracers/generic_COBALT.F90
theresa-cordero: added `c2n` conversion.

generic_tracers/generic_COBALT.F90
```diff
@@ -496,456 +508,527 @@ subroutine user_add_params
...
-    call g_tracer_add_param('zeta', cobalt%zeta, 0.05)                                   ! dimensionless
-    call g_tracer_add_param('par_adj',cobalt%par_adj, 0.83)                              ! dimensionless
-    call g_tracer_add_param('gamma_irr_aclm', cobalt%gamma_irr_aclm, 1.0 / sperd)        ! s-1
-    call g_tracer_add_param('gamma_irr_mem_dp',cobalt%gamma_irr_mem_dp,0.1/sperd)        ! s-1
-    call g_tracer_add_param('gamma_mu_mem', cobalt%gamma_mu_mem, 1.0 / sperd)            ! s-1
-    call g_tracer_add_param('ml_aclm_efold', cobalt%ml_aclm_efold, 2.5)                  ! dimensionless
-    call g_tracer_add_param('zmld_ref', cobalt%zmld_ref, 10.0)                           ! m
-    call g_tracer_add_param('densdiff_mld', cobalt%densdiff_mld, 0.03)                   ! kg m-3
-    call g_tracer_add_param('irrad_day_thresh', cobalt%irrad_day_thresh, 1.0 )           ! watts m-2
+    call get_param(param_file, "generic_COBALT", "alpha_Di_hl", phyto(DIAZO)%alpha_hl,  "alpha_Di_hl", units="g C g chl-1 s-1", &
```
theresa-cordero, line 566: @charliestock  irradiance rescaling has been added here.

generic_tracers/generic_COBALT.F90
```diff
@@ -496,456 +508,527 @@ subroutine user_add_params
...
+    call get_param(param_file, "generic_COBALT", "alpha_Lg_ll", phyto(LARGE)%alpha_ll,  "alpha_Lg_ll", units="g C g chl-1 s-1", &
+                   default= 0.8e-5, scale = micromolQpersec2W )! g C g Chl-1 sec-1 (W m-2)-1
+    call get_param(param_file, "generic_COBALT", "alpha_Md_ll", phyto(MEDIUM)%alpha_ll, "alpha_Md_ll", units="g C g chl-1 s-1", & 
+                   default= 1.6e-5, scale = micromolQpersec2W )! g C g Chl-1 sec-1 (W m-2)-1
+    call get_param(param_file, "generic_COBALT", "alpha_Sm_ll", phyto(SMALL)%alpha_ll,  "alpha_Sm_ll", units="g C g chl-1 s-1", &
+                   default= 3.2e-5, scale = micromolQpersec2W )! g C g Chl-1 sec-1 (W m-2)-1
+
+    call get_param(param_file, "generic_COBALT", "kappa_eppley",        cobalt%kappa_eppley,            "kappa_eppley",        units="deg C-1",         default= 0.063)                   ! deg C-1
+    call get_param(param_file, "generic_COBALT", "P_C_max_Di_hl", phyto(DIAZO)%P_C_max_hl, "P_C_max_Di_hl", units="day-1", &
+                   default= 0.6, scale = I_sperd ) ! s-1
```
theresa-cordero, line 585: note: where variables were divided by seconds-per-day the scale factor should be `1.0/sperd` because it is multiplied. I called this `I_sperd` for "inverse seconds-per-day" but we could have just as easily defined it as `dpers`

generic_tracers/generic_COBALT.F90
```diff
@@ -496,456 +508,527 @@ subroutine user_add_params
...
     !
-    call g_tracer_add_param('agg_Sm',phyto(SMALL)%agg,0.05*1e6 / sperd)            ! s-1 (mole N kg)-1
-    call g_tracer_add_param('agg_Di',phyto(DIAZO)%agg,  0.0    / sperd)           ! s-1 (mole N kg)-1
-    call g_tracer_add_param('agg_Lg',phyto(LARGE)%agg,0.25*1e6 / sperd)            ! s-1 (mole N kg)-1
-    call g_tracer_add_param('agg_Md',phyto(MEDIUM)%agg,0.10*1e6 / sperd)           ! s-1 (mole N kg)-1
-    call g_tracer_add_param('frac_mu_stress_Sm',phyto(SMALL)%frac_mu_stress,0.25)  ! none
-    call g_tracer_add_param('frac_mu_stress_Di',phyto(DIAZO)%frac_mu_stress,0.25)  ! none
-    call g_tracer_add_param('frac_mu_stress_Lg',phyto(LARGE)%frac_mu_stress,0.25)  ! none
-    call g_tracer_add_param('frac_mu_stress_Md',phyto(MEDIUM)%frac_mu_stress,0.25) ! none
+    call get_param(param_file, "generic_COBALT", "agg_Sm",           phyto(SMALL)%agg,  "agg_Sm", units="day-1(mol N kg)-1", &
```
theresa-cordero, line 712: add conversion from micromoles to moles.

generic_tracers/generic_COBALT.F90
```diff
@@ -496,456 +508,527 @@ subroutine user_add_params
...
-    call g_tracer_add_param('beta_fescav',cobalt%beta_fescav, 2.5e9/spery )                  ! sec-1 (mole ndet kg-1)-1
-    call g_tracer_add_param('remin_eff_fedet',cobalt%remin_eff_fedet, 0.25)                  ! unitless
-    call g_tracer_add_param('io_fescav',cobalt%io_fescav, 10.0 )                             ! watts m-2
-    call g_tracer_add_param('kfe_eq_lig_ll',cobalt%kfe_eq_lig_ll, 1.0e12)                    ! mol lig-1 kg
-    call g_tracer_add_param('kfe_eq_lig_hl',cobalt%kfe_eq_lig_hl, 1.0e9)                     ! mol lig-1 kg
+    call get_param(param_file, "generic_COBALT", "felig_bkg",          cobalt%felig_bkg,           "felig_bkg",         units="mol Fe kg-1", default= 0.5e-9)                          ! mol Fe kg-1
+    call get_param(param_file, "generic_COBALT", "felig_2_don",        cobalt%felig_2_don,         "felig_2_don",       units="mol lig kg-1", default= 0.5e-3)                          ! mol lig mol N-1
+    call get_param(param_file, "generic_COBALT", "fe_2_n_sed",         cobalt%fe_2_n_sed,          "fe_2_n_sed",        units="mol Fe kg-1", default= 100.0e-5 * 106 / 16)            ! mol Fe mol N-1
```
theresa-cordero, line 931: missed a c2n here

generic_tracers/generic_COBALT.F90
```diff
@@ -496,456 +508,527 @@ subroutine user_add_params
...
+    call get_param(param_file, "generic_COBALT", "hp_phi_det",  cobalt%hp_phi_det,  "hp_phi_det",   units="unitless", default=0.35)        ! dimensionless
     !
     !----------------------------------------------------------------------
     ! Iron chemistry
     !----------------------------------------------------------------------
     !
-    call g_tracer_add_param('felig_bkg', cobalt%felig_bkg, 0.5e-9)                          ! mol Fe kg-1
-    call g_tracer_add_param('felig_2_don', cobalt%felig_2_don, 0.5e-3)                          ! mol lig mol N-1
-    call g_tracer_add_param('fe_2_n_sed', cobalt%fe_2_n_sed, 100.0e-5 * 106 / 16)            ! mol Fe mol N-1
-    call g_tracer_add_param('ffe_sed_max', cobalt%ffe_sed_max, 170.0/1.0e6/sperd)            ! mol Fe m-2 s-1
```
theresa-cordero, line 861: is the 1/1.0e6 here correctly reinterpreted as 1/micromol2mol ?

generic_tracers/generic_COBALT.F90
```diff
@@ -496,456 +508,527 @@ subroutine user_add_params
...
+    call get_param(param_file, "generic_COBALT", "o2_min",           cobalt%o2_min,            "o2_min",           units="mol O2 kg-1   ", default= 0.8e-6 )                     ! mol O2 kg-1
+    call get_param(param_file, "generic_COBALT", "k_o2_nit",         cobalt%k_o2_nit,          "k_o2_nit",         units="mol O2 kg-1   ", default= k_o2_nit)                    ! mol O2 kg-1
+    call get_param(param_file, "generic_COBALT", "o2_min_nit",       cobalt%o2_min_nit,        "o2_min_nit",       units="mol O2 kg-1   ", default= o2_min_nit )                 ! mol O2 kg-1
+    call get_param(param_file, "generic_COBALT", "kappa_remin",       cobalt%kappa_remin,      "kappa_remin",      units="deg C-1       ", default= 0.063 )                      ! deg C-1
+    call get_param(param_file, "generic_COBALT", "remin_ramp_scale", cobalt%remin_ramp_scale,  "remin_ramp_scale", units="m             ", default= 50.0 )                       ! m
+    call get_param(param_file, "generic_COBALT", "rpcaco3",          cobalt%rpcaco3,           "rpcaco3",          units="mol N mol Ca-1", default= 0.070/12.0*16.0/106.0*100.0) ! mol N mol Ca-1
```
theresa-cordero, line 962: it looks like these parameters have a 1/c2n along with some other ratios. We could change them, but I think they are fines as is with a `do_not_read` flag.

generic_tracers/generic_COBALT.F90
theresa-cordero: added do_not_read=true for these variables.

generic_tracers/generic_COBALT.F90
```diff
@@ -496,456 +508,582 @@ subroutine user_add_params
...
-    call g_tracer_add_param('swim_max',zoo(1)%swim_max,100.0/sperd)      ! max swimming (m sec-1)
-    call g_tracer_add_param('swim_max',zoo(2)%swim_max,500.0/sperd)   ! max swimming (m sec-1)
-    call g_tracer_add_param('swim_max',zoo(3)%swim_max,2000.0/sperd)   ! max swimming (m sec-1)
+    call get_param(param_file, "generic_COBALT", "imax_smz", zoo(1)%imax, "imax_smz", units="day-1", & 
+                   default= 0.8*1.42, scale = I_sperd ) ! s-1
+    call get_param(param_file, "generic_COBALT", "imax_mdz", zoo(2)%imax, "imax_mdz", units="day-1", & 
+                   default= 0.57, scale = I_sperd ) ! s-1
+    call get_param(param_file, "generic_COBALT", "imax_lgz", zoo(3)%imax, "imax_lgz", units="day-1", & 
+                   default= 0.23, scale = I_sperd ) ! s-1
+    call get_param(param_file, "generic_COBALT", "ki_smz",           zoo(1)%ki,                "ki_smz",           units="mol N kg-1", default=1.25e-6)                        ! moles N kg-1
```
theresa-cordero, line 773: @charliestock @yichengt900 I updated these parameters to have unique names as YC pointed out. I used the smz, mdz, and lgz rather than 1, 2, and 3 based on the other parameters and to be more clear.

charliestock: Thanks Theresa.

generic_tracers/cobalt_types.F90
charliestock: Looks good!

generic_tracers/generic_COBALT.F90
charliestock: Looks good.

generic_tracers/generic_COBALT.F90
charliestock: Both approaches sound like good ideas to me.  We can stick with I_sperd unless someone has a reason to prefer dpers.

charliestock (approved):
Thank you for your work on this pull request Theresa.  While there is still work to do on the parameter descriptions/units and deciding which parameters we should include "do_not_read" for, this falls mainly on my plate and would be best addressed through subsequent pull requests.  Thus, once we confirm that the pull request doesn't change answers, I think we should accept it and work off that version.

I did have two question on the units.  Is the scaling factor reapplied when creating the output file such that the resulting parameters are interpretable both at the input and output stage?

Also, am I right in thinking that our intent is to list the units in their input/output units rather than the units they are converted to when the model runs? 

@yichengt900, if there is a way to tag the version before this change?  Perhaps that would help those currently running simulations to avoid any issues as we perform additional testing?

yichengt900:
@theresa-morrison, unfortunately, your last commit broke the automated testing again. I will have to take a look to determine why the tests are not happy with the latest changes.

theresa-cordero:
@yichengt900 it looks like the variables that use "do_not_read" are being set to 0.0, I think that is probably causing the issue with the testing.

yichengt900:
@theresa-morrison, thanks. I think this makes sense because I keep getting `FATAL: NaN in input field of reproducing_EFP_sum(_2d)` in the automated testing with your last commit. Also, I noticed that once you introduced `do_not_read`, it changed the baseline answers. I suggest we revert the changes related to `do_not_read` and test it again.

yichengt900 (approved):
@theresa-morrison, I reverted your last commit, and now this PR passes the automated testing. I noticed there are still unresolved comments from @charliestock, but I think we can address them in a separate PR since we already have significant changes in this one. I will approve this PR and merge it soon. I will also conduct a longer test runs using NWA12 test case.

@charliestock, once this PR is merged, you can start reviewing the parameter section. I have also created a [release tag](https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/releases/tag/COBALTv3_202401) for the current stable version of COBALTv3. I recommend users continue using this release tag for their production runs until the new parameter inputs are finalized.

PR #65, opened 2024-06-03
Added units and descriptions to O2 sat and Schmidt number calculations

charliestock:
This pull request adds descriptions and units to the first thirty or so parameters using the formatting from Theresa's new parameter input interface.  These are mainly the parameters to determine the O2 saturation and CO2 and O2 Schmidt numbers for air-sea exchange calculations.  

In the latter case, I uncovered a few more details on the origin of these parameter choices and have put these in the comments.  In particular, the W92 setting is actually a modified version of W92 that was provided to John Dunne by Wanninkhof to ensure the relationship was robust from -2 to 40 degrees.  These were provided in an e-mail that John forwarded to me.

I was also able to remove one parameter, "NKML", that seemed to be unused.  

I tested the simulations in the 1D water column and it ran successfully after adding the following lines to the namelist:

 &cobalt_input_nml
        parameter_filename = 'COBALT_input',
                                                 'COBALT_override' /

and adding blank COBALT_input and COBALT_override files to the directory.  We probably want to add these as standard entries/files for future runs.  Thanks to @theresa-morrison for her guidance in addressing this.

### DIFF
--- a/generic_tracers/generic_COBALT.F90
+++ b/generic_tracers/generic_COBALT.F90
@@ -399,67 +399,97 @@ contains
     call g_tracer_start_param_list(package_name)
     call get_param(param_file, "generic_COBALT", "init", cobalt%init, "init", default=.false.)
 
-    call get_param(param_file, "generic_COBALT", "htotal_scale_lo", cobalt%htotal_scale_lo, "htotal_scale_lo", units=" ", default=0.01)
-    call get_param(param_file, "generic_COBALT", "htotal_scale_hi", cobalt%htotal_scale_hi, "htotal_scale_hi", units=" ", default=100.0)
+    call get_param(param_file, "generic_COBALT", "htotal_scale_lo", cobalt%htotal_scale_lo, &
+                   "scaling factor for initializing carbon chemistry solver", units=" ", default=0.01)
+    call get_param(param_file, "generic_COBALT", "htotal_scale_hi", cobalt%htotal_scale_hi, &
+                   "scaling factor for initializing carbon chemistry solver", units=" ", default=100.0)
 
-    !  Rho_0 is used in the Boussinesq
-    !  approximation to calculations of pressure and
-    !  pressure gradients, in units of kg m-3.
-    call get_param(param_file, "generic_COBALT", "RHO_0", cobalt%Rho_0, "RHO_0", units=" ", default=1035.0)
-    call get_param(param_file, "generic_COBALT", "NKML" , cobalt%nkml,  "NKML" , units=" ", default=1)
-    !-----------------------------------------------------------------------
-    !       coefficients for O2 saturation
-    !----------------------------------------------------------------------
-    call get_param(param_file, "generic_COBALT", "a_0", cobalt%a_0, "a_0", units=" ", default=2.00907)
-    call get_param(param_file, "generic_COBALT", "a_1", cobalt%a_1, "a_1", units=" ", default=3.22014)
-    call get_param(param_file, "generic_COBALT", "a_2", cobalt%a_2, "a_2", units=" ", default=4.05010)
-    call get_param(param_file, "generic_COBALT", "a_3", cobalt%a_3, "a_3", units=" ", default= 4.94457)
-    call get_param(param_file, "generic_COBALT", "a_4", cobalt%a_4, "a_4", units=" ", default= -2.56847e-01)
-    call get_param(param_file, "generic_COBALT", "a_5", cobalt%a_5, "a_5", units=" ", default= 3.88767)
-    call get_param(param_file, "generic_COBALT", "b_0", cobalt%b_0, "b_0", units=" ", default= -6.24523e-03)
-    call get_param(param_file, "generic_COBALT", "b_1", cobalt%b_1, "b_1", units=" ", default= -7.37614e-03)
-    call get_param(param_file, "generic_COBALT", "b_2", cobalt%b_2, "b_2", units=" ", default= -1.03410e-02 )
-    call get_param(param_file, "generic_COBALT", "b_3", cobalt%b_3, "b_3", units=" ", default= -8.17083e-03)
-    call get_param(param_file, "generic_COBALT", "c_0", cobalt%c_0, "c_0", units=" ", default= -4.88682e-07)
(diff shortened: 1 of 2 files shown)


yichengt900 (approved):
Thanks @charliestock. Changes look good to me.  This PR passed automated testing and is ready to merge.

generic_tracers/generic_COBALT.F90
```diff
@@ -399,67 +399,98 @@ subroutine user_add_params(param_file)
...
-    !-----------------------------------------------------------------------
+    call get_param(param_file, "generic_COBALT", "htotal_scale_lo", cobalt%htotal_scale_lo, &
+                   "scaling factor for initializing carbon chemistry solver", units=" ", default=0.01)
+    call get_param(param_file, "generic_COBALT", "htotal_scale_hi", cobalt%htotal_scale_hi, &
+                   "scaling factor for initializing carbon chemistry solver", units=" ", default=100.0)
+
+    call get_param(param_file, "generic_COBALT", "RHO_0", cobalt%Rho_0, "reference density", &
+                   units="kg m-3", default=1035.0)
+    !------------------------------------------------------------------------------------------------------------------
+    ! Coefficients for O2 saturation based on Garcia and Gordon, 1992.  Oxygen solubility in seawater: Better fitting
```
andrew-c-ross, line 410: It looks like these are the coefficients for Garcia and Gordon's fit to only the Benson and Krause data (first columns of Table 1). Is it worth updating to the combined fit?

generic_tracers/generic_COBALT.F90
```diff
@@ -5095,7 +5126,9 @@ subroutine generic_COBALT_update_from_source(tracer_list,Temp,Salt,rho_dzt,dzt,h
       endif !}
     enddo; enddo ; enddo  !} i,j,k
 
-    ! O2 saturation
+    ! Calculate the oxygen saturation using the relationships of Garcia and Gordon, 1992. Oxygen solubility in
```
andrew-c-ross, line 5130: Garcia and Gordon equation 8 has a A3 * T_s^2 term that is missing below. Is this their mistake or ours?

andrew-c-ross (changes requested):
Just noted two things to double check before this is merged. Both things existed before this PR and aren't changed in it, though.

generic_tracers/generic_COBALT.F90
andrew-c-ross: Screenshot_20240605_100246

andrew-c-ross: Seems like their mistake. For their check at 10 degC and 35 ppt, the COBALT code as is gives 6.31476641, which matches the expected 6.315. 

```fortran
program main
    implicit none
    real o2sat
    real, parameter :: Temp = 10
    real, parameter :: Salt = 35
    real, parameter :: a_0 = 2.00907
    real, parameter :: a_1 = 3.22014
    real, parameter :: a_2 = 4.05010
    real, parameter :: a_3 = 4.94457
    real, parameter :: a_4 = -2.56847e-01
    real, parameter :: a_5 = 3.88767
    real, parameter :: b_0 = -6.24523e-03
    real, parameter :: b_1 = -7.37614e-03
    real, parameter :: b_2 = -1.03410e-02 
    real, parameter :: b_3 = -8.17083e-03
    real, parameter :: c_0 = -4.88682e-07
    real :: sal, tt, tkb, ts, ts2, ts3, ts4, ts5

    sal = min(42.0,max(0.0,Salt))
    tt = 298.15 - min(40.0,max(0.0,Temp))
    tkb = 273.15 + min(40.0,max(0.0,Temp))
    ts = log(tt / tkb)
    ts2 = ts  * ts
    ts3 = ts2 * ts
    ts4 = ts3 * ts
    ts5 = ts4 * ts

    o2sat = exp(a_0 +a_1*ts + a_2*ts2 +a_3*ts3 +a_4*ts4 +a_5*ts5 + (b_0 +b_1*ts +b_2*ts2 +b_3*ts3 +c_0*sal)*sal)
    write(*, *) o2sat
end program main
```

andrew-c-ross: Check value also matches if I use the COBALT code with the new coefficients from the paper.

```fortran
real, parameter :: a_0 = 2.00856
real, parameter :: a_1 = 3.224
real, parameter :: a_2 = 3.99063
real, parameter :: a_3 = 4.80299
real, parameter :: a_4 = 9.78188e-01
real, parameter :: a_5 = 1.71069
real, parameter :: b_0 = -6.24097e-03
real, parameter :: b_1 = -6.93498e-03
real, parameter :: b_2 = -6.90358e-03
real, parameter :: b_3 = -4.29155e-03
real, parameter :: c_0 = -3.1168e-07
```

generic_tracers/generic_COBALT.F90
charliestock: I will take this up with John Dunne and Jasmin John to discuss whether there was a rationale for this.  For now, I will add a bit to the comment to indicate that it uses the fit to the Benson and Krause data.  I'll add a new commit with this addition shortly.

generic_tracers/generic_COBALT.F90
charliestock: That was indeed their mistake, not ours.  Thanks for the careful checks!

charliestock:
Added the Benson and Krause fit reference to the comment and sent a note to JPD and JGJ.  Let's forge ahead with merging!

andrew-c-ross (approved):

yichengt900:
Thanks, @charliestock and @andrew-c-ross . Looks like now everyone is happy with this PR. I will merge it now.

andrew-c-ross:
@charliestock I did some more searching and confirmed that the Benson and Krause fit is what should have been used for OCMIP-2: https://web.archive.org/web/20051017101155fw_/http://www.ipsl.jussieu.fr/OCMIP/phase2/simulations/Biotic/boundcond/o2sato.f

charliestock:
Thanks Andrew.  I see they didn't offer any information on why they chose Benson and Krause.  John knows Orr well, so I pinged him to see if he could get any insight before deciding on whether a change is warranted.

PR #67, opened 2024-06-10
cleaned up plankton foodweb code

charliestock:
This pull request cleans up the code associated with zooplankton and higher predator grazing.  This involved:

1) Removing legacy code from COBALTv2 that had just been commented out
2) Removing code associated with an experimental diapause approximation that was not being used.
3) Removing an experimental "self grazing" setting that was not being used.
4) Improved commenting throughout.

The removal of the experimental diapause formulation means that the diagnostic tracer "irr_mem_dp" is no longer needed.  Several diagnostics were also removed in the clean up, but none of these seems to appear in the diagnostic table.  

I have tested the code and it runs.  This should not change answers relative to the last version, and it ultimately removed about 100 lines of code.

### DIFF
--- a/generic_tracers/generic_COBALT.F90
+++ b/generic_tracers/generic_COBALT.F90
@@ -3180,7 +3180,7 @@ contains
     ! Zooplankton feeding is parameterized with observed allometric (i.e., size-dependent) feeding rates and predator-
     ! prey linkages (i.e., Hansen, B.W. et al., 1994; Hansen, P.J., et al. 1997).  Feeding relationships are based on
     ! simple saturating (Holling Type 2) relationships when there is a single prey type.  A density dependent switching
-    ! response, however, is included when multiple prey types are present (Stock et al., 2008).  A small "refuge
+    ! response, however, is included when multiple prey types are present (Stock et al., 2008).  A small "refuge"
     ! concentration has also been included as an extra safeguard against negative values and as a reflection of the 
     ! paradigm that Baas Becking's hypothesis that "Everything is everywhere, but the environment selects".  Predation
     ! by higher predators (e.g., planktivorous fish) is modeled in a manner analogous to zooplankton, but assuming that
--- a/generic_tracers/generic_COBALT.F90
+++ b/generic_tracers/generic_COBALT.F90
@@ -653,8 +653,6 @@ contains
     call get_param(param_file, "generic_COBALT", "par_adj",             cobalt%par_adj,                 "par_adj",             units="",            default= 0.83)  ! dimensionless
     call get_param(param_file, "generic_COBALT", "gamma_irr_aclm",      cobalt%gamma_irr_aclm,          "gamma_irr_aclm",      units="day-1", &
                    default= 1.0, scale = I_sperd ) ! s-1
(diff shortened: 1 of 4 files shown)


generic_tracers/generic_COBALT.F90
```diff
@@ -3189,31 +3177,49 @@ subroutine generic_COBALT_update_from_source(tracer_list,Temp,Salt,rho_dzt,dzt,h
     !
     ! 3.1 Plankton foodweb dynamics: consumption by zooplankton and higher predators
     !
+    ! Zooplankton feeding is parameterized with observed allometric (i.e., size-dependent) feeding rates and predator-
+    ! prey linkages (i.e., Hansen, B.W. et al., 1994; Hansen, P.J., et al. 1997).  Feeding relationships are based on
+    ! simple saturating (Holling Type 2) relationships when there is a single prey type.  A density dependent switching
+    ! response, however, is included when multiple prey types are present (Stock et al., 2008).  A small "refuge
```
yichengt900: missing closing quotation mark: "refuge"

generic_tracers/generic_COBALT.F90
yichengt900, line 1: @charliestock , Thanks for opening this PR. Overall the changes are good. Since you are cleaning up the plankton section,  I'd like to ask your opinion on the[ vertical movement section for both phytoplankton and zooplankton](https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/70c41f66ae23ed8469ed6c05a162f4c9c0b7fc8f/generic_tracers/generic_COBALT.F90#L3615-L3668). Do we want to keep this capability, or would we prefer to revert it back to COBALTv2? If we decide to keep it, I suggest we fix the [parameter name issue](https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/70c41f66ae23ed8469ed6c05a162f4c9c0b7fc8f/generic_tracers/generic_COBALT.F90#L811-L816) for `swim_max` in this PR too.

Also I can confirm this PR does not change answers in `NWA12-RT` case.

jessluo (approved):
looks good to me!

generic_tracers/generic_COBALT.F90
charliestock: Hi YC - I think we can remove the zooplankton movement for now, but we need to keep the phytoplankton sinking.  @jessluo will eventually have a more comprehensive & correct movement formulation to add.  

I can do a follow up pull request for the zooplankton swimming tomorrow, but perhaps we can pull in the foodweb changes today.  This was mainly what I was hoping to cover in tomorrow's meeting, so a clean code would be helpful.

charliestock:
Great, thanks @yichengt900 and @jessluo - why don't we blend this in today and the next pull requests can build from this.

best,
Charlie

generic_tracers/generic_COBALT.F90
yichengt900: @charliestock Thanks. It sounds like a plan to me. I don't want to be annoyingly picky, but the `swim_max` issue generates a bunch of annoying warning messages during model runtime. I'll be happy to see it getting fixed in the next PR. Other than that, this PR looks good, except for one super tiny comment I left above. I will approve this PR once it is resolved.

yichengt900 (approved):

Issue #71, opened 2024-06-17
Removing the "ocmip2" option?


charliestock:
In a prior meeting, the potential removal of the "ocmip2" option for carbon system calculations was raised.  This is technically the default setting for carbon system calculations but it is never used as far as I am aware.   John Dunne's comments also seemed to imply that it was clearly inferior to mocsy.  I have sent an e-mail to John to confirm his view on this.  If it does prove to be the case, there seems to be a significant opportunity to simplify and shorten the code here.  A few thoughts:

1. We could retain https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/dev/cefi/generic_tracers/FMS_ocmip2_co2calc.F90 or move it to FMS_ocmip2_co2calc.F90.archive to preserve this structure for some period of time (I suspect several other routines may still use it.

2. We could create FMS_co2calc.F90 from this routine by reducing it to a "mocsy" interface.

3. We could remove the co2_calc option.

4. The following sections of generic_COBALT.F90 could be removed and/or simplified:
https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/b1dd5ecc7a85398eee8f63861381eabf75926edd/generic_tracers/generic_COBALT.F90#L101-L105
https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/b1dd5ecc7a85398eee8f63861381eabf75926edd/generic_tracers/generic_COBALT.F90#L274-L280
https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/b1dd5ecc7a85398eee8f63861381eabf75926edd/generic_tracers/generic_COBALT.F90#L2353-L2394
https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/b1dd5ecc7a85398eee8f63861381eabf75926edd/generic_tracers/generic_COBALT.F90#L3876-L3900
https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/b1dd5ecc7a85398eee8f63861381eabf75926edd/generic_tracers/generic_COBALT.F90#L5972-L6000

Thoughts?  @yichengt900 @jessluo @gabyneg?

yichengt900:
PR #72 solved this issue. Close now.

PR #72, opened 2024-06-18
Simplify and clean up co2 calculation code

yichengt900:
This PR addresses issue #71. The following modifications have been made:

1. Renamed `FMS_ocmip2_co2calc.F90` to `FMS_ocmip2_co2calc.F90.archive`.
2. Created `FMS_co2calc.F90`, which contains the necessary routines/interface for `mocsy` co2 calculation.
3.  Simplified `generic_COBALT.F90`, `generic_BLING.F90`, and `generic_abiotic.F90`, so they now work properly with the new `FMS_co2calc.F90`.

The only thing I have not removed yet is the `co2_calc` option in the namelist. Removing `co2_calc` would affect the model configurations because we would have to take it out of `input.nml` to prevent the model from complaining about an unknown namelist.

This PR should not change answers and it has passed both the 1D case. and `NWA12-COBALT` RT  case.


charliestock:
This is great @yichengt900   Do you think a 1 year NWA12 test that looks at the DIC and/or air-sea exchange before and after this is needed to ensure that no subtle change has been introduced?

yichengt900:
@charliestock, good point. I conducted a one-year NWA12 test to check if the DIC results from this PR are identical to the results from the current `dev/cefi` branch. The good news is they are identical (see the plots below). I think this PR is now ready for review.

`ALK, DIC, omega from current `dev/cefi` branch`:
alk_dic_omega

`ALK, DIC, omega from this PR`:
alk_dic_omega

charliestock (approved):
Hello @yichengt900.  Thank you for cleaning this up.  I think it will help us avoid many issues moving forward.  I have checked the changes and they all seemed sensible.  The fact that the simulated DIC and alkalinity fields were bitwise reproducible even with a full year also provides high confidence.  I have approved the request, but had 2 thoughts to consider before finalizing (or for a later pull request):

1) I will run the details by John Dunne offline.
2) Are we sure that FMS_ocmip2_co2calc.F90 isn't called anywhere else?  For example, in the air-sea exchange codes?

For the latter, Eric Stofferahn mentioned that there are a number of "ocmip2" linked codes that Brandon Reichl was modifying for his work.  It sounded like these were unrelated to the CO2 calculation that we are working on here, but could we double check that there are no cross-linkages?

Thanks again!
Charlie

yichengt900:
@charliestock,

Thank you. Regarding your 2 question, I checked the FMS air-sea exchange codes and found some ocmip2-linked codes. However, none of them seem to be related to `FMS_ocmip2_co2calc.F90` directly. I will keep this PR open for a while to ensure everything is okay before we merge it.

generic_tracers/generic_COBALT.F90
```diff
@@ -271,13 +265,12 @@ subroutine generic_COBALT_register(tracer_list)
       write (stdoutunit,*) trim(note_header), 'Simulating radiocarbon'
     endif
 
-    if (trim(co2_calc) == 'ocmip2') then
-      write (stdoutunit,*) trim(note_header), 'Using FMS OCMIP2 CO2 routine'
-    else if (trim(co2_calc) == 'mocsy') then
+    if (trim(co2_calc) == 'mocsy') then
```
jessluo, line 268: Do we need this moving forward, if we're not providing multiple options for the co2 calculation?

jessluo (approved):
This looks good to me. I just left one quick question but otherwise no additional suggestions for this PR.

generic_tracers/generic_COBALT.F90
jessluo: Oh, I see there is already an issue open regarding this: [https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/issues/71](url)

yichengt900: @jessluo, thank you! Yes, this is the first step for #71, simplifying the current CO2 calculation module. We are keeping `co2_calc` in this PR but will consider removing it in the future if we offer only one option to users.

PR #78, opened 2024-07-01
Remove vertdiff_M routines and make vertdiff_G just vertdiff

### LINKED ISSUE #48: Cleanup: remove generic_tracer_vertdiff_M and g_tracer_vertdiff_M?
I don't see any uses of these two routines in the current generic tracer code, and I think they're a vestige from the CMIP5 era? As far as I can see all code is now using generic_tracer_vertidff_G, which we could also rename to just generic_tracer_vertdiff.

andrew-c-ross:
This resolves #48 by removing everything related to vertdiff_M and renaming vertdiff_G to just vertdiff.

For this PR to work, MOM6 and ocean_BGC need to make the switch at at the same time (MOM6 PR: NOAA-CEFI-Regional-Ocean-Modeling/MOM6#11).


yichengt900:
@andrew-c-ross , Thanks! I reviewed the code changes and it lks good to me. I left my small concern [here](https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/MOM6/pull/11#issuecomment-2200723353) and perhaps we can discuss this during tomorrow's COBALTv3 dev meeting.

yichengt900:
@nikizadehgfdl, I am adding you to keep you in the loop. My understanding is that once we decide to push this change back to the GFDL main repo, it should not impact the existing experiment runs since we normally use specific tags for the experiments in xml. However, any additional thoughts you might have would be greatly appreciated.

andrew-c-ross:
(We decided to revert the renaming of the vertdiff_G routines so that MOM6 doesn't need to be changed, with a longer-term vision of unifying the MOM6 and generic tracer vertdiff routines)

yichengt900 (approved):
Thanks @andrew-c-ross,  everything looks good. Approve.

PR #79, opened 2024-07-05
Ballast and remin/clean code

charliestock:
This pull request cleans up the mineral ballasting, remineralization of sinking detritus and iron scavenging sections of COBALTv3.  It gets rid of one redundant local variable (z_remin_ramp) and introduces a parameter for the augmentation of scavenging when free iron exceeds fe solubility (fast_fescav_fac).  

For fast_fescav_fac have been using a factor of 2 for this in the regional runs, but previous global coarse resolution runs used a factor of 10.  I think this parameter is somewhat scale dependent.  I used a higher value in prior global runs to try and erode coastal iron signals that were propagating too far into the ocean interior, probably because shelf residence times were under-estimated with coarse resolution.  Higher resolution addressed this problem and a factor of 2 seems to work better.  I have thus made 2 the default.

Most of this pull request is commenting and clean up.  Perhaps most significantly, the comments now include a detailed description of the ligand binding derivation used to calculate the free iron (feprime).  This involves solving a system of three equations and results in a quadratic equation for feprime.  The comments should now provide enough information for users to be able to understand and verify the derivation.

One thing I did not have time to do yet is to enhance all the variable definitions associated with this pull request.  I will handle this in a follow on pull request soon.    

This pull request should not change answers!

### DIFF
--- a/generic_tracers/generic_COBALT.F90
+++ b/generic_tracers/generic_COBALT.F90
@@ -2251,7 +2251,6 @@ contains
     integer, dimension(:,:), Allocatable :: k_bot, kblt
     real, dimension(:), Allocatable   :: tmp_irr_band
     real, dimension(:,:), Allocatable :: rho_dzt_100,rho_dzt_200,rho_dzt_bot,sfc_irrad
-    real, dimension(:,:,:), Allocatable :: z_remin_ramp
     real,dimension(1:NUM_ZOO,1:NUM_PREY) :: ipa_matrix,pa_matrix,ingest_matrix
     real,dimension(1:NUM_PREY) :: hp_ipa_vec,hp_pa_vec,hp_ingest_vec
     real,dimension(1:NUM_PREY) :: prey_vec,prey_p2n_vec,prey_fe2n_vec,prey_si2n_vec
@@ -2366,8 +2365,8 @@ contains
          co2star=cobalt%co2_csurf(:,:), alpha=cobalt%co2_alpha(:,:), &
          pCO2surf=cobalt%pco2_csurf(:,:), &
          co3_ion=cobalt%f_co3_ion(:,:,k), &
-         omega_arag=cobalt%omegaa(:,:,k), &
-         omega_calc=cobalt%omegac(:,:,k))
+         omega_arag=cobalt%omega_arag(:,:,k), &
+         omega_calc=cobalt%omega_calc(:,:,k))
 
     do k = 2, nk
        do j = jsc, jec ; do i = isc, iec  !{
@@ -2388,8 +2387,8 @@ contains
             zt=cobalt%zt(:,:,k),                          &
                                 !OUT
             co3_ion=cobalt%f_co3_ion(:,:,k), &
-            omega_arag=cobalt%omegaa(:,:,k), &
-            omega_calc=cobalt%omegac(:,:,k))
+            omega_arag=cobalt%omega_arag(:,:,k), &
+            omega_calc=cobalt%omega_calc(:,:,k))
     enddo
 
     call g_tracer_set_values(tracer_list,'htotal','field',cobalt%f_htotal  ,isd,jsd)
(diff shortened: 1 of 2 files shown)


generic_tracers/generic_COBALT.F90
```diff
@@ -3934,116 +4004,157 @@ subroutine generic_COBALT_update_from_source(tracer_list,Temp,Salt,rho_dzt,dzt,h
...
 !
 !-------------------------------------------------------------------------------------------------
-! 7: Sedimentary/coastal fluxes/transformations
+! 5: Sedimentary/coastal fluxes/transformations
 !-------------------------------------------------------------------------------------------------
 !
 
     !
-    ! Iceberg and "Coastal" iron and other nutrient input
+    ! Iceberg and "Coastal" iron and other nutrient input.  The
```
jessluo: is the "The" extra here?

generic_tracers/generic_COBALT.F90
```diff
@@ -4207,9 +4316,9 @@ subroutine generic_COBALT_update_from_source(tracer_list,Temp,Salt,rho_dzt,dzt,h
           ! Iron flux from the sediment
           !
 
-          ! iron from sediment (Elrod)
-          !cobalt%ffe_sed(i,j) = cobalt%fe_2_n_sed * cobalt%f_ndet_btf(i,j,1)
-          ! iron from sediment (Dale, 2015)
+          ! Iron from sediment (Dale, 2015).  The maximum release from the sediment is set by ffe_sed_max.  The
+          ! hyperbolic tanget requires the flux of carbon to the sediments (as mmoles m-2 day-1) in the numerator
```
jessluo: "tangent"

generic_tracers/generic_COBALT.F90
```diff
@@ -3934,116 +4004,157 @@ subroutine generic_COBALT_update_from_source(tracer_list,Temp,Salt,rho_dzt,dzt,h
...
+    ! COBALT uses a single ligand complexation model for iron scavenging onto detritus (e.g., Archer and Johnson, 2000).
+    ! Ther binding strength of the ligand, however, is modulated between weak high-light (kfe_eq_hl) and strong low-
+    ! light limits (kfelig_ll) to mimic the weakening effect that oxygen free radicals have on iron binding in well-lit
+    ! waters (Fan, 2008).  The weakest binding is at light levels greater than io_fescav = 10 watts m-2.  Values decline 
+    ! to the strongest low-light limit at 0.01 watts m-2.
+    !
+    ! The ligand concentration includes a background concentration (felig_bkg) and an additional amount proportional
+    ! to dissolved organic matter (felig_2_don).  When the free iron (feprime) exceeds solubility limits defined as
+    ! a function of temperature and salinity according to Liu and Millero (2002), scavenging is increased by the factor
+    ! fast_scav_fac to mimic rapd precipitation.  In coarse resolution global simulations, fast_scav_fac was generally
```
jessluo: "rapid"

generic_tracers/generic_COBALT.F90
```diff
@@ -3934,116 +4004,157 @@ subroutine generic_COBALT_update_from_source(tracer_list,Temp,Salt,rho_dzt,dzt,h
...
-!--------------------------------------------------------------------------------------------
-! 6: Iron
-!--------------------------------------------------------------------------------------------
-!
-
+    ! 
+    ! 4.5: Iron scavenging onto detritus
+    !  
+    ! COBALT uses a single ligand complexation model for iron scavenging onto detritus (e.g., Archer and Johnson, 2000).
+    ! Ther binding strength of the ligand, however, is modulated between weak high-light (kfe_eq_hl) and strong low-
```
jessluo: "The"

jessluo (commented):
Still need some time to go through this in detail. But to start with, here are just a couple typos I caught in my first read through.

generic_tracers/generic_COBALT.F90
```diff
@@ -3855,65 +3913,76 @@ subroutine generic_COBALT_update_from_source(tracer_list,Temp,Salt,rho_dzt,dzt,h
...
+    ! The saturation state dependence is capped with the parameter caco3_sat_max.   
     do k = 1, nk ; do j = jsc, jec ; do i = isc, iec   !{
+        ! Pteropods are assumed to be the primary aragonite shell formers.  Pteropods fall into the medium and large
+        ! zooplankton groups within COBALT.  Production of aragonite detritus is thus linked to the consumption of
+        ! medium and large zooplankton by zooplankton and higher predators, and the proportion of the material consumed
+        ! that ends up as detritus (i.e., phi_det).
         cobalt%jprod_cadet_arag(i,j,k) = (zoo(2)%jzloss_n(i,j,k)*zoo(3)%phi_det + &
                        (zoo(2)%jhploss_n(i,j,k) + zoo(3)%jhploss_n(i,j,k))*cobalt%hp_phi_det)* &
                        cobalt%ca_2_n_arag*min(cobalt%caco3_sat_max, max(0.0,cobalt%omega_arag(i,j,k) - 1.0)) + epsln
+        ! Forams and coccolithophores are assumed to the primary calcite shell formers.  Forams fall into the small
```
yichengt900: are assumed to **be** the primary

generic_tracers/generic_COBALT.F90
```diff
@@ -6563,8 +6672,6 @@ subroutine user_allocate_arrays
     allocate(cobalt%jnamx(isd:ied, jsd:jed, 1:nk)) ; cobalt%jnamx=0.0
     allocate(cobalt%omega_arag(isd:ied, jsd:jed, 1:nk))   ; cobalt%omega_arag=0.0
     allocate(cobalt%omega_calc(isd:ied, jsd:jed, 1:nk))   ; cobalt%omega_calc=0.0
-    allocate(cobalt%omegaa(isd:ied, jsd:jed, 1:nk))       ; cobalt%omegaa=0.0
```
yichengt900, line 6566: We should also remove `omegaa` and `omegac` from `cobalt_types.F90` if they are not used:

https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/b5203ce9cb4f52c9674696bbd0e5d672676a087e/generic_tracers/cobalt_types.F90#L718-L719

generic_tracers/generic_COBALT.F90
```diff
@@ -3934,116 +4004,157 @@ subroutine generic_COBALT_update_from_source(tracer_list,Temp,Salt,rho_dzt,dzt,h
...
-       !
-       ! Under sub-oxic conditions
-       !
+
+       ! Calculate remineralization under anaerobic conditions
        else !}{
           cobalt%jremin_ndet(i,j,k) = cobalt%gamma_ndet * cobalt%o2_min / &
                (cobalt%k_o2 + cobalt%o2_min)* &
-               cobalt%f_no3(i,j,k) / (phyto(SMALL)%k_no3 + cobalt%f_no3(i,j,k))* &
+               cobalt%f_no3(i,j,k) / (cobalt%k_no3_denit + cobalt%f_no3(i,j,k))* &
```
yichengt900, line 4051: Although the new code makes sense to me, it changes our baseline in the NWA12-RT case. I checked COBALTv2 and found that we also used small phytoplankton's `k_no3`. @charliestock just wants to confirm if this is the change we want.

generic_tracers/generic_COBALT.F90
charliestock: Thanks for catching these Jessica.

generic_tracers/generic_COBALT.F90
charliestock: Agreed.  I will add this along with the various fixes above to another commit.

generic_tracers/generic_COBALT.F90
charliestock: Does it?  I think cobalt(SMALL)%k_no3 and cobalt%k_no3_denit have the same value.  We just introduced k_no3_denit to be explicit that this was a different paramater.

charliestock: I stand corrected.  These parameters are different.  This will change answers, but in a minimal way.  The parameter is only has a significant effect when both oxygen and nitrate have been exhausted which rarely, if ever, happens.  Even in such cases, the impact will be small.

In general, I suggest we remove uses of the phytoplankton parameters beyond their explicit definition.  It looks like we used phyto(SMALL)%k_no3, for example, for the anammox dependence on NO3.  Thus, changing phyto(SMALL)%k_no3 risks unintentionally changing multiple factors.  I think we should adopt k_no3_denit for denitrification and define k_no2_anammox for anammox and set both to 1e-6.  What do you folks think?

@yichengt900 @jessluo

generic_tracers/generic_COBALT.F90
charliestock: Yes.  Removed.  I'll have a full cleanup/commenting of the sediment in next week's work.

charliestock:
This new commit fixes typos in the comments (thanks Jessica), eliminates omegaa and omegac from cobalt_types.F90, and retains the k_no3_denit change in the water column remin (which will change answers very slightly), and introduces k_no3_amx.  The last of these changes will have no impact since anammox isn't turned on but, line k_no3_denit, establishes a clear separation between this parameter and the phytoplankton uptake parameters.  @jessluo @yichengt900

yichengt900:
@charliestock, I like the idea of removing the uses of the phytoplankton parameters beyond their explicit definition. The new commit looks good to me. I will wait a bit for others' comments.

gabyneg (approved):
The changes, and the added descriptions look good to me, and make it easier to follow.

generic_tracers/generic_COBALT.F90
jessluo: Yes, I completely agree. It was just a quick shorthand to use phyto(SMALL)%k_no3 but doing so poses risks where a user may come in and change what they think is just the small phytoplankton nitrate half saturation constant and end up changing values for other processes as well.

yichengt900 (approved):
@charliestock, overall the changes look good, and the comments are straightforward and easy to understand. Although this PR will change our baselines due to the introduction of the new runtime parameter `k_no3_denit` with a slightly different value (we also added `k_no3_amx`, but it is not yet in use), we expect the impact to be minor. Please approve this PR.

jessluo (approved):
Looks good to me!

Issue #85, opened 2024-07-18
Geothermal heating must be on for COBALT to work


andrew-c-ross:
This one has been around for a while. COBALT uses the internal (geothermal) heat argument to calculate an iron flux:

https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/1bc8466335573dc62cf89eb69ec982339b0413b2/generic_tracers/generic_COBALT.F90#L4326

However if you trace it back far enough, internal heat is a null pointer by default that only gets allocated if `DO_GEOTHERMAL = True` in the MOM_input:
https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/MOM6/blob/3624434308f2863985353091cd02ee274cd18b7b/src/core/MOM.F90#L2802

So we probably need to check if internal_heat is associated to avoid the segfault that occurs if DO_GEOTHERMAL = False, and either fail with a descriptive warning or assume geothermal heating is 0 (currently we assume it is 0 anyways in the regional models).

PR #86, opened 2024-07-18
Make internal_heat optional so it works without geothermal heating

andrew-c-ross:
As in #85 , if DO_GEOTHERMAL = False, COBALT would segfault. With this PR, it will now check if `internal_heat` is a null pointer (by making it an optional argument and checking if it is present) and assume 0 internal heating if it is null. 

Careful checking of my work here would be appreciated since pointers can be devious little creatures.


yichengt900 (approved):
@andrew-c-ross, the changes look good to me. I tested the 1D case locally by turning off geothermal heating, and this patch works as expected.

charliestock:
Thanks for addressing this Andrew, YC.  Much appreciated.

PR #88, opened 2024-07-31
Update interfaces to pass a photoacclimation MLD from MOM6

theresa-cordero:
This PR changes the intefaces between MOM6 and COBALT to allow a mixed layer depth to be passed from MOM6. The new argument is optional, and if present used instead of calculating the photoacclimation depth in COBALT.


yichengt900:
Hi @theresa-morrison , thank you for opening this draft PR. I understand that it has a few dependencies and may require further refinement. I will begin by reviewing it and running several tests, including a 1D test case and a 5-year simulation using the NWA12 domain. Will keep you updated on the progress.

theresa-cordero:
Hi @yichengt900 thanks!  I was planning to talk through the code with @charliestock tomorrow to make sure it looks like it will do what the COBALT group wants before starting the 5 year test.

yichengt900:
@theresa-morrison @andrew-c-ross, and @charliestock:

The preliminary results (5yr NWA run) with the new interface (MLD pre-calculated in MOM6 and passed to COBALT) are quite promising. First, the new `MLD_003` diagnostics (set `HREF_FOR_MLD=10.0`) gives a slight better bias when compared to the dBM 2022 (see figures below):

**new `MLD003` when `HREF_FOR_MLD=10.0`:**
mld003_eval

**Org `MLD003`:**
mld003_eval

The domain and regional surface `chl` also look ok to me:
 domain_chl
regional_chl

yichengt900:
As a reference, below are the surface `chl` values from our most recent bi-weekly regression testing. As you can see, the results are quite close to those using the MOM6-calculated photoacclimation MLD (figures above):

domain_chl


regional_chl

charliestock:
Thanks @yichengt900 and @theresa-morrison, this looks great.  I am tempted to make the leap to this improved approach sooner than later.  Let's discuss whether there should be any final checks on this when we get together tomorrow.  If not, let's move ahead with the merge.  If we think a few more checks are necessary, we can set them in motion before the weekend and confirm things next week.  Thank you both!

yichengt900:
Ok here are some new updates from the `1993-2019` long term NWA12 runs:

**Regional Chl:**
regional_chl

**Domain Chl:**
domain_chl

**GoM Hypoxia :**
hypoxic_area_from_daily

**MLD003:**
mld003_eval_new

Although it’s not an exact apple-to-apple comparison to the results presented in Andrew's paper, the overall consistency is within the expected range. One observation from the domain chlorophyll plots is that the discrepancy along the OBC appears to be larger in the new run. This could potentially be related to the shorter OBC update frequency I used in this test. Hopefully, the new BGC nudging PR proposed by @theresa-morrison could mitigate this issue.

Overall, the new interface changes look promising. I would suggest retaining the COBALT-MLD calculation for now until we’re all comfortable with the new interface and MOM6-calculated MLD for various regions. We can still remove the `eq_of_state` argument from the MOM6 side, which should not impact this PR.

generic_tracers/generic_COBALT.F90
```diff
@@ -2694,58 +2696,65 @@ subroutine generic_COBALT_update_from_source(tracer_list,Temp,Salt,rho_dzt,dzt,h
     ! captures mixing on time scale of 1 to a few days.
     ! de Boyer-Montegut reference:  https://doi.org/10.1029/2004JC002378
     !
-    do j = jsc, jec ; do i = isc, iec   !{
+    if (present(photo_acc_dpth)) then
+      call mpp_error(WARNING, "Using a photoacclimation MLD in COBALTv3 that was calculated in "//&
```
theresa-cordero: @yichengt900 
It would be great if there was a check here to see if photo_acc_dpth is a fixed value, but I am not sure how to do that. Any ideas?

generic_tracers/generic_tracer.F90
```diff
@@ -523,7 +523,7 @@ subroutine generic_tracer_source(Temp,Salt,rho_dzt,dzt,hblt_depth,ilb,jlb,tau,dt
     real,                      optional , intent(in) :: sosga ! global avg. sea surface salinity
     real, dimension(ilb:,jlb:),  intent(in) :: geolat 
     type(EOS_type),              intent(in) :: eqn_of_state
-
+    real, dimension(ilb:,jlb:), optional, intent(in) :: photo_acc_dpth
```
theresa-cordero, line 526: Once the EOS can be removed we can make this a non-optional argument.

generic_tracers/generic_COBALT.F90
yichengt900: @theresa-morrison , do you mean to check if `photo_acc_dpth` does not contain any NaN values?

theresa-cordero: no I want to check "photo_acc_dpth=0.0" everywhere - or at least that it is a constant, if not 0.0? Does that make sense?

charliestock: Hi Theresa, YC - Theresa and I discussed this a bit last Friday.  In my mind, using a constant photoacclimation depth in COBALT is effectively an error.  Theresa mentioned that she wanted to keep the option to not calculate it for those folks that are using other BGC modules (or none), which seems like a good idea, but you shouldn't use COBALT without a time varying photoacclimation depth, so like the idea of passing zeros and providing an error message so the user knows there is obviously something wrong with the model and has an idea on how to solve.

yichengt900: @theresa-morrison, this should be relatively straightforward to implement. If `photo_acc_dpth` is a constant array, the value at any position should be the same across the array. We can try the following trick:
```
logical :: is_all_same
is_all_same = all(photo_acc_dpth == photo_acc_dpth(isc,jsc))
if (is_all_same) then
 ! Include your error message here for COBALTv3
endif
```
If you still encounter any issues while coding this, I'm more than happy to help.

theresa-cordero: thanks! `all` is new to me and exactly what I was looking for. I will check that this works and add it to the PR

theresa-cordero: This has been added and works as expected.

theresa-cordero: I should add, contrary to our expectations, COBALT does not crash with a photoacclimation MLD of 0.0 in a short 24 hour run.

generic_tracers/generic_COBALT.F90
```diff
@@ -2702,58 +2705,68 @@ subroutine generic_COBALT_update_from_source(tracer_list,Temp,Salt,rho_dzt,dzt,h
     ! captures mixing on time scale of 1 to a few days.
     ! de Boyer-Montegut reference:  https://doi.org/10.1029/2004JC002378
     !
-    do j = jsc, jec ; do i = isc, iec   !{
+    pha_all_same = all(photo_acc_dpth == photo_acc_dpth(isc,jsc))
```
yichengt900: @theresa-morrison, I now understand why the regression testing crashed. Since `photo_acc_dpth` isn't being passed at this stage (as the MOM6 side hasn't been updated), the array is never declared. This leads to an out-of-bounds error when performing the `all_same` check outside the `if (present(photo_acc_dpth))` loop. To resolve this, you can move the `all_same` check inside the if loop.

generic_tracers/generic_COBALT.F90
```diff
@@ -2332,6 +2334,7 @@ subroutine generic_COBALT_update_from_source(tracer_list,Temp,Salt,rho_dzt,dzt,h
     real :: rev_angle, dec_angle, temp_arg
 
     logical ::  phos_nh3_override
+    logical ::  pha_all_same
```
yichengt900: Should we set this default to false?

generic_tracers/generic_COBALT.F90
theresa-cordero: done!

generic_tracers/generic_COBALT.F90
theresa-cordero: sure, I added default false - do you think true is better?

yichengt900: Hmm as a second thought, true might be a better option - assume it is always a fixed variable array unless we turn on calculation.

yichengt900 (approved):
@theresa-morrison, thanks for this PR and the new MOM6-calculated photo_acc MLD! The code changes look good to me. Regarding the default value of `pha_all_same`, I trust you to choose the one that aligns with your logic on the MOM6 side. Approved for now.

theresa-cordero:
Thanks, Yi-Cheng. I decided `pha_all_same=true` is a better default so there is one more commit. 
I think this PR should be ready to go and can be merged before the changes to MOM6. 
This PR alone should have no impact on the model since `photo_acc_dpth` will not be present in the call to update COBALT.

PR #89, opened 2024-08-04
Sed and coastal/clean code

charliestock:
This pull request cleans up and adds commenting to sections of the sediment code.  It also modifies the formulation of a previously used "coastal" iron source (jfe_coast).  This source of iron was used in TOPAZ and early versions of COBALT to simulate coastal iron fluxes from the vertical land face.  This was done because coarse resolution often prevented significant sediment iron fluxes from being introduced near the ocean surface in shallow waters: the grid was so coarse, even the first grid cell was far deeper than the euphotic zone in most places.  Releasing iron from the land edges provided a means of generating "island wakes" even when the model resolution prevented resolution of shallow nearshore areas.

As the resolution of our ocean model configurations has increased, we have moved away from using this "coastal" source, instead relying on the benthic fluxes that are explicitly resolved by the model.  This has worked for nearly all of our CEFI runs.  Recent Pacific Island simulations done by @gabyneg, however, suggest that there still may be a role for this parameterized flux in areas with particularly narrow shelves and steep bathymetry.  However, when I looked back at the old way the flux was formulated I found the underlying rationale lacking.  I thus coded an alternative that made the coastal flux proportional to the benthic flux that would have arisen from the organic carbon flux and O2 in the adjacent water.  Conceptually, one can think of this as a net flux that would have arisen from unresolved shallow regions within the first grid cell.  The parameter "fe_coast" becomes a constant of proportionality which we need to explore but I suspect should be ~0.1 for cases where we feel it is necessary to use this parameterization.

The default value of fe_coast is 0 and I feel it should stay this way.  With fe_coast = 0, this pull request should not change answers.  In addition to the code review and commenting, it would be great if @gabyneg could run a few test cases in her Pacific Islands domain to see if we are satisfied with the new approach.

I still have a bit more work to do with the commenting of the sediment section, but I will complete this in a subsequent pull request.


charliestock:
Hello @yichengt900, I made the small adjustments we discussed this morning.  As expected, it did not seem to fix the cryptic restart issue blocking the pull request.  Looking at the code changes, I can't imagine what might be tripping this off.  When you get a chance, could you have a look at the restarts and see if they provide any clues?

yichengt900:
Hi @charliestock,

I reviewed the differences in the restart files, and the variables showing discrepancies after the model integrated the 3600s time step (the model ran with restart for an hour integration) are listed below:

```
nccmp -dfqS RESTART/MOM.res.nc RESTART_25hrs/MOM.res.nc
Variable Group Count          Sum      AbsSum         Min          Max       Range         Mean      StdDev
htotal   /        40 -4.40209e-19 4.40209e-19 -1.1162e-20 -1.08261e-20 3.35835e-22 -1.10052e-20 9.42072e-23
co3_ion  /        40  4.17849e-15 4.17849e-15 1.02782e-16  1.05723e-16  2.9409e-18  1.04462e-16 7.98836e-19
cased    /         8  1.36514e-14 1.36514e-14 1.70643e-15  1.70643e-15           0  1.70643e-15           0
dic      /        40  8.87918e-15 8.87918e-15  2.2031e-16  2.23346e-16 3.03577e-18   2.2198e-16 8.47963e-19
alk      /        40   1.7754e-14  1.7754e-14 4.42354e-16  4.45824e-16 3.46945e-18  4.43851e-16 7.54041e-19
```  
I will investigate further.

generic_tracers/generic_COBALT.F90
```diff
@@ -4313,9 +4321,40 @@ subroutine generic_COBALT_update_from_source(tracer_list,Temp,Salt,rho_dzt,dzt,h
...
+          ! O2 level of the adjacent waters.  This then spread across the layer mass (rho_dzt(i,j,k)) to give an input
+          ! in moles Fe kg-1 sec-1. Conceptually, this can be though of as a net iron flux resulting from the fraction
+          ! of the sinking flux that would have been intercepted at shallower depths were the model resolution finer.
+          ! default value of fe_coast is 0 (i.e., only the explicitly resolved benthic flux is included).
+          !
+          ! Old Expression:
+          ! cobalt%jfe_coast(i,j,1) = cobalt%fe_coast * mask_coast(i,j) * grid_tmask(i,j,1) / &
+          !     sqrt(grid_dat(i,j))
+          !
+          do k = 1, nk !{
```
yichengt900, line 4337: Hi @charliestock,
I think I found the issue. At line 4241, we specify `k = grid_kmt(i,j)` to ensure that we use the model bottom later for the calculation in the sediment section.
https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/07b96471c2868a313ad4d56baa6f627e002d604e/generic_tracers/generic_COBALT.F90#L4241
However, the new code changes the k value and eventually affects `fcased_burial` starting at line 4418:
https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/07b96471c2868a313ad4d56baa6f627e002d604e/generic_tracers/generic_COBALT.F90#L4418C22-L4427

The solution would be to add `k = grid_kmt(i,j)` either right after line 4357 or right before line 4418, or just simply change line4426 to `grid_tmask(i,j,grid_kmt(i,j))`

yichengt900 (changes requested):
Hi @charliestock,

I think I found the issue that is breaking the reproducibility across restarts.

Please see my comment below.

generic_tracers/generic_COBALT.F90
charliestock: I agree that this is likely the issue.  We should, however, discuss the role of grid_kmt in MOM6 when we meet tomorrow.  My understanding is that grid_kmt = nk everywhere for MOM6, so I'm not sure this oversight actually changed anything.  Let's implement and find out!

charliestock:
That seemed to do the trick YC, thanks!

generic_tracers/generic_COBALT.F90
yichengt900: @charliestock, I think the main issue is that in Fortran after 'k=1,nk', the k value will be +1 over nk (76) in memory, which will be out of bound and then causing potential issue. But sure we can discuss more tomorrow.

generic_tracers/generic_COBALT.F90
```diff
@@ -4144,38 +4152,43 @@ subroutine generic_COBALT_update_from_source(tracer_list,Temp,Salt,rho_dzt,dzt,h
...
-            sqrt(grid_dat(i,j))
        cobalt%jfe_iceberg(i,j,k) = 0.0
        cobalt%jno3_iceberg(i,j,k) = 0.0
-       cobalt%jpo4_iceberg(i,j,k) = 0.0
+       cobalt%jpo4_iceberg(i,j,k) = 0.0 
     enddo; enddo; enddo  !} i,j,k
 
+    ! Calculate the bottom conditions and the fluxes to the bottom for diagnostics and benthic flux calculations.
+    ! MOM4/5 used the bottom grid cell, but MOM6 often has a number of vanishingly thin layers overlying the bottom.
+    ! Grid scale noise in these layers can occur, particularly for quantitities with large bottom fluxes.  COBALT thus
```
yichengt900: Should be `quantities`

generic_tracers/generic_COBALT.F90
```diff
@@ -4313,9 +4321,40 @@ subroutine generic_COBALT_update_from_source(tracer_list,Temp,Salt,rho_dzt,dzt,h
 
           ! Iron from sediment (Dale, 2015).  The maximum release from the sediment is set by ffe_sed_max.  The
           ! hyperbolic tangent requires the flux of carbon to the sediments (as mmoles m-2 day-1) in the numerator
-          ! and the bottom water oxygen concentration (in microMolar units) in the denominator:
+          ! and the bottom water oxygen concentration (in microMolar units) in the denominator. Note that ffe_sed_max
+          ! was converted to moles Fe m-2 sec-1 during parameter input, so ffe_sed has is in moles Fe m-2 sec-1
```
yichengt900: `has is` should be `is`

generic_tracers/generic_COBALT.F90
```diff
@@ -4313,9 +4321,40 @@ subroutine generic_COBALT_update_from_source(tracer_list,Temp,Salt,rho_dzt,dzt,h
...
                                 max(cobalt%btm_o2(i,j)*1.0e6,epsln) )
+
+          ! Additional coastal iron (Optional, default fe_coast = 0)
+          !
+          ! Coarse resolution models and/or intermediate resolution models in areas with exceptionally steep bathymetry
+          ! can under-represent coastal iron because they don't resolve shallow regions. An option to add iron through
+          ! the vertical face of the land mass has thus been included.  The flux is posed as a fraction (fe_coast) of
+          ! the sediment Fe flux (moles Fe m-2 sec-1) that would have result from the sinking organic matter flux and
+          ! O2 level of the adjacent waters.  This then spread across the layer mass (rho_dzt(i,j,k)) to give an input
+          ! in moles Fe kg-1 sec-1. Conceptually, this can be though of as a net iron flux resulting from the fraction
```
yichengt900: `thought`?

generic_tracers/generic_COBALT.F90
```diff
@@ -4249,6 +4256,7 @@ subroutine generic_COBALT_update_from_source(tracer_list,Temp,Salt,rho_dzt,dzt,h
              ! Since burial is highly uncertain and often used in global earth system simulations to balance inputs and
              ! outputs, a dimensionless scaling factor (cobalt%scale_burial) has also been included.
              fpoc_btm = cobalt%fntot_btm(i,j)*cobalt%c_2_n*sperd*1000.0
+             ! Should we use ztop?
```
yichengt900: Hmm. This depends on how we define our benthic layer. If we consider the model bottom layer as the benthic layer, then using `ztop` would be appropriate. However, if we define the benthic layer as one layer below our model's bottom grid, then `cobalt%zt` would be suitable.

yichengt900 (commented):
@charliestock , this PR looks good after the fix, I just have a few small comments.

generic_tracers/generic_COBALT.F90
```diff
@@ -4313,9 +4321,40 @@ subroutine generic_COBALT_update_from_source(tracer_list,Temp,Salt,rho_dzt,dzt,h
...
+          ! was converted to moles Fe m-2 sec-1 during parameter input, so ffe_sed has is in moles Fe m-2 sec-1
           cobalt%ffe_sed(i,j) = cobalt%ffe_sed_max * tanh( (cobalt%fntot_btm(i,j)*cobalt%c_2_n*sperd*1.0e3)/ &
                                 max(cobalt%btm_o2(i,j)*1.0e6,epsln) )
+
+          ! Additional coastal iron (Optional, default fe_coast = 0)
+          !
+          ! Coarse resolution models and/or intermediate resolution models in areas with exceptionally steep bathymetry
+          ! can under-represent coastal iron because they don't resolve shallow regions. An option to add iron through
+          ! the vertical face of the land mass has thus been included.  The flux is posed as a fraction (fe_coast) of
+          ! the sediment Fe flux (moles Fe m-2 sec-1) that would have result from the sinking organic matter flux and
```
gabyneg: change "result" to "resulted" or remove "have" in this sentence "the sediment Fe flux (moles Fe m-2 sec-1) that would have result from the sinking organic matter flux and O2 level of the adjacent waters."

generic_tracers/generic_COBALT.F90
```diff
@@ -908,15 +908,23 @@ subroutine user_add_params(param_file)
...
-    call get_param(param_file, "generic_COBALT", "fe_coast",           cobalt%fe_coast,            "fe_coast",          units="mol Fe m kg-1 s-1", default= 0.0 )                                ! mol Fe m kg-1 s-1
+    call get_param(param_file, "generic_COBALT", "ffe_geotherm_ratio", cobalt%ffe_geotherm_ratio, &
+                   "iron release per unit of geothermal heat",units="mol Fe m-2 s-1 (W m-2)-1", default= 2.0e-12)
+    call get_param(param_file, "generic_COBALT", "jfe_iceberg_ratio",  cobalt%jfe_iceberg_ratio, &
+                   "iron release per kg of ice melt", units="mol Fe kg-1 ice melt", default= 1.0e-7)
+    call get_param(param_file, "generic_COBALT", "jno3_iceberg_ratio", cobalt%jno3_iceberg_ratio, &
+                   "nitrate release per kg of ice melt",units="mol N kg-1 ice melt", default= 2.0e-6)
+    call get_param(param_file, "generic_COBALT", "jpo4_iceberg_ratio", cobalt%jpo4_iceberg_ratio, &
+                   "phosphate release per kg of ice melt",units="mol P kg-1 ice melt", default= 1.1e-7)
+    ! fe_coast is effectively the fraction of the equivalent benthic iron flux that you would have been generated by
```
gabyneg: Remove "you" after "... flux that" in "fe_coast is effectively the fraction of the equivalent benthic iron flux that you would have been generated by", and change "were it to encounter sediments" to "when encountering sediments".

generic_tracers/generic_COBALT.F90
```diff
@@ -4313,9 +4321,40 @@ subroutine generic_COBALT_update_from_source(tracer_list,Temp,Salt,rho_dzt,dzt,h
...
+
+          ! Additional coastal iron (Optional, default fe_coast = 0)
+          !
+          ! Coarse resolution models and/or intermediate resolution models in areas with exceptionally steep bathymetry
+          ! can under-represent coastal iron because they don't resolve shallow regions. An option to add iron through
+          ! the vertical face of the land mass has thus been included.  The flux is posed as a fraction (fe_coast) of
+          ! the sediment Fe flux (moles Fe m-2 sec-1) that would have result from the sinking organic matter flux and
+          ! O2 level of the adjacent waters.  This then spread across the layer mass (rho_dzt(i,j,k)) to give an input
+          ! in moles Fe kg-1 sec-1. Conceptually, this can be though of as a net iron flux resulting from the fraction
+          ! of the sinking flux that would have been intercepted at shallower depths were the model resolution finer.
```
gabyneg, line 4330: add "is" after model resolution

gabyneg (commented):
I just made a few edits on the comments, and will start working on creating a few model tests to see how this changes the Main Hawaiian island domain.

gabyneg (commented):
I set up 3 experiments with the updated `fe_coast` formulation, where fe_coast =0.001, 0.01, and 0.1. Below I show the differences in surface dissolved iron compared to a base run without any changes to fe_coast (fe_coast =0). 
Screenshot 2024-08-09 at 6 16 27 PM
(the columns represent base and differences between experiments (exp - base)). I have also looked at how surface chlorophyll 
Screenshot 2024-08-09 at 6 16 18 PM
and the deep chlorophyll maximum at station ALOHA were modified by this change. 
Screenshot 2024-08-09 at 6 21 00 PM
 (DCM doesn't change between experiments). The changes between these three experiments do not seem to look linear, so I am having trouble understanding why the smallest changes in fe_coast (0.001), seem to show the largest differences in dissolved iron along the islands. I have verified my xml and code to make sure I didn't make a mistake plotting the figures.

generic_tracers/generic_COBALT.F90
```diff
@@ -4144,38 +4152,43 @@ subroutine generic_COBALT_update_from_source(tracer_list,Temp,Salt,rho_dzt,dzt,h
...
        if (grid_kmt(i,j) .gt. 0) then !{
           cobalt%jfe_iceberg(i,j,1) = cobalt%jfe_iceberg_ratio*max(frunoff(i,j),0.0)/rho_dzt(i,j,1)
           cobalt%jno3_iceberg(i,j,1) = cobalt%jno3_iceberg_ratio*max(frunoff(i,j),0.0)/rho_dzt(i,j,1)
           cobalt%jpo4_iceberg(i,j,1) = cobalt%jpo4_iceberg_ratio*max(frunoff(i,j),0.0)/rho_dzt(i,j,1)
-          cobalt%jfe_coast(i,j,1) = cobalt%fe_coast * mask_coast(i,j) * grid_tmask(i,j,1) / &
-               sqrt(grid_dat(i,j))
        endif !}
     enddo; enddo  !} i,j
+    ! CAS: Is the necessary?
     do k = 2, nk ; do j = jsc, jec ; do i = isc, iec   !{
```
andrew-c-ross: I don't thinks so, since these are initialized with 0s [here](https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/d57dbe98a033352d316a5ccd52afa18e0ff62eb0/generic_tracers/generic_COBALT.F90#L6694:L6696)

generic_tracers/generic_COBALT.F90
charliestock: got it.

generic_tracers/generic_COBALT.F90
charliestock: thanks.

generic_tracers/generic_COBALT.F90
charliestock: yup.

generic_tracers/generic_COBALT.F90
charliestock: I know I have looked at this directly before, and maybe I'm just thinking myself in a circle, but am I right in thinking that grid_kmt is equal to the number of vertical tracer grid cells (e.g., 75 for most of our regional applications) in MOM6.  So "zt(75)" is the bottom of the 75th grid cell (i.e., the sediment)?  I just need a quick sanity check on this.

generic_tracers/generic_COBALT.F90
charliestock: Got it.  Thanks.

generic_tracers/generic_COBALT.F90
charliestock: got it.

generic_tracers/generic_COBALT.F90
charliestock: I think this one might be OK as is, but I did need a "The" after finer.

generic_tracers/generic_COBALT.F90
charliestock: Agreed.  I have deleted.  and we will confirm there are no answer changes.

charliestock: Hi Gaby.  Thank you for adding the figures and sorry it has taken me so long to look at them.  I too am a bit surprised to not see a clearer progression from the small to the largest flux, but I think this may be partly due to the prominent boundary condition noise.  I will be good to continue testing this in combination with Theresa's boundary nudging fixes.

charliestock:
It looks like this has passed checks.  Since the default fe_coast = 0 won't effect any answers, perhaps we can merge it in?

yichengt900 (approved):
Thanks, @charliestock. The changes look good. I tested them with the NWA12 regression, and the results are identical. Approved.

PR #93, opened 2024-09-03
added parameter definitions for phytoplankton growth + cleanup

charliestock:
This pull request adds parameter definitions and cleans up the units for the phytoplankton growth parameters in COBALTv3.  It covers about 25% of the parameter definitions.  It should not change answers.  It would be good, however, to revisit our discussion of the units for parameter inputs.  I think it works OK for parameters like alpha - the initial slope of the photosynthesis-irradiance curve - and rate parameters since these do not have explicit units in the variable names.  I'm a bit concerned, however, about parameters like "fe_2_n".  Entering values as fe_2_c and then converting them may create confusion.  We can discuss this further at the next doc&dev meeting, along with a few other odds and ends.

### DIFF
--- a/generic_tracers/generic_COBALT.F90
+++ b/generic_tracers/generic_COBALT.F90
@@ -536,129 +536,252 @@ contains
     ! Nutrient Limitation Parameters (phytoplankton)
     !-----------------------------------------------------------------------
     !
-    call get_param(param_file, "generic_COBALT", "k_fed_Di", phyto(DIAZO)%k_fed,            "k_fed_Di",       units="mol Fed kg-1", default=4.0e-9)  ! mol Fed kg-1
-    call get_param(param_file, "generic_COBALT", "k_fed_Lg", phyto(LARGE)%k_fed,            "k_fed_Lg",       units="mol Fed kg-1", default=2.0e-9)  ! mol Fed kg-1
-    call get_param(param_file, "generic_COBALT", "k_fed_Md", phyto(MEDIUM)%k_fed,           "k_fed_Md",       units="mol Fed kg-1", default=8.0e-10) ! mol Fed kg-1
-    call get_param(param_file, "generic_COBALT", "k_fed_Sm", phyto(SMALL)%k_fed,            "k_fed_Sm",       units="mol Fed kg-1", default=4.0e-10) ! mol Fed kg-1
-    call get_param(param_file, "generic_COBALT", "k_nh4_Lg", phyto(LARGE)%k_nh4,            "k_nh4_Lg",       units="mol NH4 kg-1", default=5.0e-8)  ! mol NH4 kg-1
-    call get_param(param_file, "generic_COBALT", "k_nh4_Md", phyto(MEDIUM)%k_nh4,           "k_nh4_Md",       units="mol NH4 kg-1", default=2.0e-8)  ! mol NH4 kg-1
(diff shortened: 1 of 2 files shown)


generic_tracers/generic_COBALT.F90
```diff
@@ -536,129 +536,252 @@ subroutine user_add_params(param_file)
...
+    call get_param(param_file, "generic_COBALT", "alpha_Lg_ll", phyto(LARGE)%alpha_ll, &
+                   "Chl-a specific initial slope of photosynth-irrad curve, low-light adapted large phytoplankton", &
+                   units="g C g Chl-1 (micromol quanta m-2)-1", default=0.8e-5, scale = micromolQ2Joule)
+    call get_param(param_file, "generic_COBALT", "alpha_Md_ll", phyto(MEDIUM)%alpha_ll, &
+                   "Chl-a specific initial slope of photosynth-irrad curve, low-light adapted medium phytoplankton", &
+                   units="g C g Chl-1 (micromol quanta m-2)-1",default=1.6e-5, scale = micromolQ2Joule)
+    call get_param(param_file, "generic_COBALT", "alpha_Sm_ll", phyto(SMALL)%alpha_ll, &
+                   "Chl-a specific initial slope of photosynth-irrad curve, low-light adapted small phytoplankton", &
+                   units="g C g Chl-1 (micromol quanta m-2)-1", default=3.2e-5, scale = micromolQ2Joule)
+    ! Phytoplankton Maximum photosythetic rate parameters
```
yichengt900: Should be photosynthetic?

generic_tracers/generic_COBALT.F90
```diff
@@ -536,129 +536,252 @@ subroutine user_add_params(param_file)
...
+                   default=4.0e-6, scale=c2n)
+    call get_param(param_file, "generic_COBALT", "k_fe_2_n_Sm",phyto(SMALL)%k_fe_2_n, & 
+                   "internal iron quota half-saturation for small phytoplankton growth", units="mol Fe (mol N)-1", &
+                   default=2.0e-6, scale=c2n)
+    call get_param(param_file, "generic_COBALT", "fe_2_n_max_Di", phyto(DIAZO)%fe_2_n_max, &
+                   "maximum internal iron quota for diazotrophs",  units="mol Fe (mol N)-1", &
+                   default=500.0e-6, scale=c2n)
+    call get_param(param_file, "generic_COBALT", "fe_2_n_max_Lg", phyto(LARGE)%fe_2_n_max, &
+                   "maximum internal iron quota for large phytoplankton",  units="mol Fe (mol N)-1", &
+                   default=500.0e-6, scale=c2n)
```
theresa-cordero, line 612: After the discussion at the COBALT Doc & Dev meeting today we thought it would be better to have the input in mol Fe (mol N)-1 which would make the default here  `default=c2n*500.0e-6`. 

We should check if this impacts the ability to exactly represent the defaults in the parameter file.

yichengt900: @theresa-morrison, thanks for the note (I was not able to join until the last 20 minutes). If this is the case, I would recommend that we open an issue and address it through a separate PR

yichengt900 (approved):
@charliestock, thanks for this PR adding detailed parameter definitions. I only fixed a small typo; everything else looks good to me. We will address the remaining parameters and the unit conversion issue in a separate PR. Approved.

PR #96, opened 2024-09-16
changed fe_2_n parameters so defaults were entered as Fe:N and not Fe:C

charliestock:
The previous code was set up to receive the default values of a number of iron parameters as moles Fe:mole C, but the parameter names and ultimate values are "fe_2_n".  We discussed this issue at the COBALT doc & dev meeting on 9/4 and I believe the consensus was that this was likely to cause confusion, particularly because the units are explicit in the parameter name.  This pull requests thus sets the default values of these parameters to units of moles Fe:mole N.  I have, however, maintained the explicit multiplication by "c2n" in the default setting so users are aware of what the equivalent Fe:C values are.


yichengt900 (approved):
@charliestock, thank you for opening this PR to address issue #96. The code changes look good. While it’s a bit of a missed opportunity not to utilize the `scale` capability provided by the `get_param` function, I understand your point about minimizing potential user confusion. Approved.

PR #97, opened 2024-09-18
Plankton stoichiometry params/clean code

charliestock:
This pull request cleans up the parameter definitions for the plankton stoichiometry and eliminated a few "static" parameters that are no longer needed.  I did find a typo when going through these but I am not sure if it is answer changing.    The Si:N ratios for the diatoms on lines 817-818 of the original code were set as follows:

call get_param(param_file, "generic_COBALT", "si_2_n_max_Lg",    phyto(LARGE)%si_2_n_max,      "si_2_n_max_Lg",   units="mol Si mol N-1", default= 3.0)                  ! mol Si mol N-1
    call get_param(param_file, "generic_COBALT", "si_2_n_max_Lg",    phyto(MEDIUM)%si_2_n_max,     "si_2_n_max_Lg",   units="mol Si mol N-1", default= 1.0)                 ! mol Si mol N-1

The line for the large phytoplankton was correct.  The line for the medium, however, still used  "si_2_n_max_Lg" as the third get_param entry, though the 4th entry is correct and I think this is the one that actually impacts what the model uses.  I checked si_2_n_max_Lg in the COBALT parameter output and it had the right value there.  Anyway, if the pull request changes answers, it will likely be reflected in minor changes to the silicate, but I am not sure that it will.

The other changes should be straight forward and improve the overall documentation.

### DIFF
--- a/generic_tracers/cobalt_types.F90
+++ b/generic_tracers/cobalt_types.F90
@@ -82,7 +82,6 @@ module cobalt_types
      real ::  alpha_hl          !< Chlorophyll a-specific initial slope of the photosynthesis-irradiance curve high level (g C g Chl-1 sec-1 (W m-2)-1)
      real ::  alpha_ll          !< Chlorophyll a-specific initial slope of the photosynthesis-irradiance curve low level (g C g Chl-1 sec-1 (W m-2)-1)
      real ::  fe_2_n_max        !< Maximum iron to nitrogen ratio (mol Fe mol N-1)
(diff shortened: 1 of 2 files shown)


yichengt900 (approved):
@charliestock, thanks for this PR and catching that typo in `si_2_n_max_Md!` We got lucky this time because, even with the typo, the default value for `phyto(MEDIUM)%si_2_n_max` still worked. Just an excellent example of how important it is to keep our code tidy! 😄

Issue #102, opened 2024-12-03
Flang and Clang generate file open issues in COBALT


jbrzensk:
When building with flang and clang, and running the OM4 1D example case, an error is generated on open_param_file:
```
 MOM_MOSAIC domain decomposition
whalo =    3, ehalo =    3, shalo =    3, nhalo =    3
  X-AXIS =    4
  Y-AXIS =    4

FATAL: open_param_file: Error opening 'COBALT_input'.
```
Some explicit messages were added to the MOM6_File_parser, including warnings before they were fatal.

```
WARNING: open_param_file: Error opening 'COBALT_input'.
 Error opening code:           207 .
WARNING: Error opening message: 'file is already connected to another unit'.
FATAL: open_param_file: Error opening 'COBALT_input'.
```

FORTRAN allows concurrent access to a file, but it is based on the compiler and the system. 
GCC is very tolerant of this, while FLANG is not. 
IFORT is somewhere in the middle depending on the system. 
It’s generally better to avoid this practice when possible. For better resource management, a module that opens a file should also close it once it’s done.

I fixed the issue by removing the extra close_param_file in cobalt_param_doc.F90:
https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/0de1ab230a12999eb139763f40e56920a204d639/generic_tracers/cobalt_param_doc.F90#L8

And adding it to generic_COBALT.F90
https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/0de1ab230a12999eb139763f40e56920a204d639/generic_tracers/generic_COBALT.F90#L157

And then calling it at the end of user_add_tracers, before line 2422

https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/0de1ab230a12999eb139763f40e56920a204d639/generic_tracers/generic_COBALT.F90#L2415-L2422

I can push my branch but I do not have permissions on your repo, or someone can make the three changes.

yichengt900:
Hi @jbrzensk, Thank you for identifying the bug! We greatly appreciate your effort—this truly reflects the spirit of community collaboration.

We haven’t tested the new COBALT with FLANG, but your assessment sounds correct to me. The only suggestion I have is to consider adding the `close_param_file` call at line 1642:

[...]

PR #103, opened 2024-12-04
Close param_file for issue 102

jbrzensk:
I closed the param file in generic_COBALT.F90  to address compile/runtime errors with possible concurrent access to the param file from COBALT, referenced in issue #102


yichengt900:
@jbrzensk, thank you for submitting this PR. The changes appear solid, and the updates have successfully passed the 1D regression test. @theresa-morrison, it would be great if you could have a second eye on it as well to ensure we haven’t missed anything.

yichengt900:
@jbrzensk, after discussing with our MOM6 expert @theresa-morrison, we agree that this is a solid PR, and the `close_param_file` call is indeed necessary and will help to address #102. Additionally, we noticed duplicate `user_add_params` calls that will require further cleanup. I'll address that in a separate PR. Thank you again, @jbrzensk, for helping make the COBALT codebase more robust!

yichengt900 (approved):

theresa-cordero (approved):

PR #106, opened 2024-12-05
additional parameter cleanup

charliestock:
This pull request includes additional clean up and commenting of the parameter specification code.  The changes include improvements to the sediment dynamics parameters and nitrification and anammox parameters.  Part of this involved removing some (but not all) of the legacy namelist parameters from Fabian.  This pull request should not change answers, but some of the changes did require subtle shifts in the way parameters were calculated.  I don't think these will change anything at higher decimal places, but we'll take a closer look if it does not pass the answer change check.


charliestock:
Note that this pull request partly addresses issue #20.  I will try and weed out the final anomalous namelist use elements in a subsequent pull request.

generic_tracers/generic_COBALT.F90
```diff
@@ -1521,70 +1501,130 @@ subroutine user_add_params(param_file)
...
-    call get_param(param_file, "generic_COBALT", "o2_min_nit",       cobalt%o2_min_nit,        "o2_min_nit",       units="mol O2 kg-1   ", default= o2_min_nit )                 ! mol O2 kg-1
+                   "nitrate half-saturation for anammox", units="mol NO3 kg-1", default= 1.0e-6)
     !
     !-----------------------------------------------------------------------
     ! Miscellaneous
     !-----------------------------------------------------------------------
     !
-    call get_param(param_file, "generic_COBALT", "tracer_debug",  cobalt%tracer_debug, "tracer_debug", default=.false.)
+    ! Unused?
+    call get_param(param_file, "generic_COBALT", "tracer_debug",  cobalt%tracer_debug, &
```
yichengt900, line 1633: It appears this is not currently being used. We can retain it for now and reconsider removing it later if no one finds it essential. However, I believe it could still be valuable for ongoing code development.

yichengt900:
@charliestock, thanks for opening this PR. The changes look good to me (no typos this time! 😊). These updates have been tested with 1D regression tests and `NWA12`, `NEP10`, and `OM4p25 COBALT` tests, confirming no impact on the baseline results. However, the `COBALT_parameter_doc.*` will be changed because some parameters were renamed, and some old namelist parameters were moved into `get_param` calls.  The parameter values remain consistent and unchanged.

yichengt900 (approved):

charliestock:
I think the small parameter name change is OK.  We had a parameter called "o2_min_amx" that was really not a minimum but the maximum o2 that could be present for anammox to occur, so it should be "o2_max_amx".  Do you agree @jessluo?

generic_tracers/generic_COBALT.F90
charliestock: It could be useful, but I'm not sure what the legacy use was.  Let's leave it for now and consider the potential use before deleting.

jessluo:
> I think the small parameter name change is OK. We had a parameter called "o2_min_amx" that was really not a minimum but the maximum o2 that could be present for anammox to occur, so it should be "o2_max_amx". Do you agree @jessluo?

Yes, I think this is appropriate. I think the reason it was `o2_min_amx` was because originally we had it in the same section as denitrification, but then we moved it out on its own.. and gave it an _amx tag so that this oxygen threshold could be used as a tuning parameter to get the overall global anammox values in the right ballpark. Anyway, this change makes a lot of sense and is totally appropriate.

jessluo (approved):
Looks good to me!

PR #107, opened 2024-12-16
Move MOM_generic_tracer.F90 to ocean_BGC

theresa-cordero:
Move MOM_generic_tracer out of MOM6 to allow changes to the interface between MOM6 and generic tracer packages to be tailored to each group of generic tracer packages.

This should not change answers.

PR #108, opened 2025-01-02
Added V3 version MOM_generic_tracer.F90

yichengt900:
As discussed, we are relocating `MOM_generic_tracer.F90` (the BGC driver) from MOM6 to COBALT to provide greater flexibility for future development. This PR does not modify any baselines and will function correctly once this CEFI MOM6 [PR](https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/MOM6/pull/26) is merged. This PR is nearly identical to #107, with a minor cleanup of `MOM_generic_tracer.F90` to ensure it passes the MOM6 format check, in case we decide to replace the old version of `MOM_generic_tracer.F90` in the current GFDL MOM6 repository. This PR must be tested with the latest dev/cefi [MOM6 update](https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/MOM6/pull/26).

### DIFF
--- a/generic_tracers/MOM_generic_tracer.F90
+++ b/generic_tracers/MOM_generic_tracer.F90
@@ -13,1137 +13,1137 @@ module MOM_generic_tracer
 #define _ALLOCATED allocated
 #endif
 
-  ! ### These imports should not reach into FMS directly ###
-  use field_manager_mod, only: fm_string_len
+! ### These imports should not reach into FMS directly ###
+use field_manager_mod, only: fm_string_len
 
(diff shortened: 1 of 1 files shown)


theresa-cordero:
@yichengt900 
This looks fine to me. Since we are changing every line in the file, could we also correct the indenting? nearly every line after ~16 is indented two extra spaces. It's a minor thing, but since we  are changing the blame on all the lines anyway it seems like a good time to do this clean up step.

yichengt900:
@theresa-morrison, thank you for the suggestion! I’ve updated the indenting accordingly.

theresa-cordero (approved):

Issue #110, opened 2025-01-03
ffetot_btm diagnostic is empty


jessluo:
The diagnostic field `ffetot_btm` is currently not being written out correctly; this field is empty (zeros) in current runs. In looking through the code, I am not entirely sure what is missing. A comparison with a field that is being written out (`fsitot_btm`) seems to show that both diagnostics are declared and written out similarly. 

```
./cobalt_reg_diag.F90:    vardesc_temp = vardesc("ffetot_btm","Total Fe sinking flux to bottom",'h','1','s','mol m-2 s-1','f')
./cobalt_reg_diag.F90:    cobalt%id_ffetot_btm = register_diag_field(package_name, vardesc_temp%name, axes(1:2),&
./cobalt_send_diag.F90:       used = g_send_data(cobalt%id_ffetot_btm,   cobalt%ffetot_btm,             &
./cobalt_types.F90:          ffetot_btm,&
./cobalt_types.F90:          id_ffetot_btm    = -1,       &
./generic_COBALT.F90:          cobalt%ffetot_btm(i,j) = cobalt%f_fedet_btf(i,j,1) + cobalt%f_fedi_btf(i,j,1) + &
./generic_COBALT.F90:          !cobalt%b_fed(i,j) = - cobalt%ffetot_btm(i,j)
./generic_COBALT.F90:    allocate(cobalt%ffetot_btm(isd:ied, jsd:jed))         ; cobalt%ffetot_btm=0.0
./generic_COBALT.F90:    deallocate(cobalt%ffetot_btm)
```

```
./cobalt_reg_diag.F90:    vardesc_temp = vardesc("fsitot_btm","Total Si sinking flux to bottom",'h','1','s','mol m-2 s-1','f')
./cobalt_reg_diag.F90:    cobalt%id_fsitot_btm = register_diag_field(package_name, vardesc_temp%name, axes(1:2),&
./cobalt_send_diag.F90:       used = g_send_data(cobalt%id_fsitot_btm,   cobalt%fsitot_btm,             &
./cobalt_types.F90:          fsitot_btm,&
./cobalt_types.F90:          id_fsitot_btm    = -1,       &
./generic_COBALT.F90:          cobalt%fsitot_btm(i,j) = cobalt%f_sidet_btf(i,j,1) + cobalt%f_silg_btf(i,j,1) + &
[...]

Issue #109, opened 2025-01-03
Improved global budget diagnostics and checks


charliestock:
COBALT includes conservation checks for sources and sinks, and prior analysis showed that the existing diagnostics could be used to close the budget with high accuracy.  The budget "noise" generated with the existing diagnostics, however, remains larger than machine precision might suggest.  This was particularly true for quantities with prominent benthic fluxes, such as iron.

@yichengt900, @jessluo and @theresa-morrison and I have been discussing this over the past week or so and have been working to address it.  This issue is intended to provide a forum for discussion and bring in others that may be interested.  We have identified two possible contributors, which I have summarized as I understand them, below:

1) Water column integrals are calculated in "update_from_source".  Theresa's flow diagram shows that this occurs before the tridiagonal solver.  While this step will not change the water column integral will not change for quantities that are only diffused, there will be a net change for particles that sink, or those that have a surface/bottom sources entered through the tri-diagonal.  A snapshot of the water column integral at the end of year, for example, would be missing 1 time step of surface/bottom sources.

2) The bottom flux from the previous time step is saved and propagated to the next time step where it generally determines the benthic sources (minus removals from things like burial and denitrifications).  The bottom flux/benthic source combination are both saved for that time step, even though the benthic flux is technically from the prior time steps call to the tridiagonal solver.  If we wanted all flux diagnostics to align perfectly with the time step on which they occurred, we should save the bottom flux diagnostic after the tridiagonal solver.

When looking at annual budgets like we have done in the past, or any time the time step is short relative to the time scale of the budget calculation, one would expect both 1) and 2) to yield very small discrepancies.  This, and the fact that the discrepancies are a bit more prominent for quantities with strong benthic components, is consistent with the behavior that we have seen in the past.  I am hoping this issue will help us to get rid of this noise.  Please chime in! 

Charlie

Tagging @nikizadehgfdl for his input.

yichengt900:
#110 is related to this issue.

charliestock:
I wanted to unpack both issues a bit further before discussing tomorrow, starting with the bottom flux (2 above)

The flux of sinking tracers to the bottom is calculated in the tri-diagonal solver (g_tracer_vertdiff_G) where the accumulated flux is saved as g_tracer%btm_reservoir:

[...]

PR #112, opened 2025-01-14
Enable ffetot_btm diagnostic

jessluo:
This pull request enables the `ffetot_btm` diagnostic, and addresses issue #110. While this is not a fix for the issue of global integrals and bottom diagnostics being saved at different time steps (see issue #109), this at least enables budget calculations for iron without having to write out four additional diagnostics, `ffesm_btm`, `ffemd_btm`,  `ffelg_btm`, and `ffedet_btm`.

### DIFF
--- a/generic_tracers/generic_COBALT.F90
+++ b/generic_tracers/generic_COBALT.F90
@@ -2647,6 +2647,8 @@ contains
     !! handled separately later using a relationship based on Dale et al., 2015. 
     call g_tracer_get_values(tracer_list,'fedi','btm_reservoir',phyto(DIAZO)%ffe_btm,isd,jsd)
     phyto(DIAZO)%ffe_btm = phyto(DIAZO)%ffe_btm/dt
+    call g_tracer_get_pointer(tracer_list,'fedi_btf','field',temp_field)
+    temp_field(:,:,1) = phyto(DIAZO)%ffe_btm(:,:)
     call g_tracer_set_values(tracer_list,'fedi','btm_reservoir',0.0)
     used = g_send_data(phyto(DIAZO)%id_ffe_btm,phyto(DIAZO)%ffe_btm, &
     model_time, rmask = grid_tmask(:,:,1),&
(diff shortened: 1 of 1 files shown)


yichengt900 (commented):
@jessluo, thank you for opening this PR. The change looks good to me. However, this update only addresses the missing pointer for `fedet_btf`. Ideally, we need to include `fedi_btf`, `fesm_btf` (which currently does not sink), `femd_btf`, and `felg_btf` to reflect the correct `ffetot_btm`:
https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/154cd599514fda31837d08143fb0680dbbcee18b/generic_tracers/generic_COBALT.F90#L4671-L4672

While the `detritus` term is likely the dominant contributor, meaning `ffetot_btm` may not be significantly off if we neglect the other terms, it could be beneficial to include the other components for more accurate values. @charliestock, do you have any thoughts on this?

jessluo:
Hi @yichengt900 - I'm confused by your comment. The other `fe*_btf` terms were not commented out in the same way as `fedet_btf`. When I run with this change, I now have a non-zero `ffetot_btm`, which is similar to but does not exactly equal to `ffesm_btm+ffemd_btm+ffelg_btm+ffedet_btm`. But I thought that is most likely due to the timestep issue between the vertical integrals and the bottom diagnostics.

jessluo:
Following @yichengt900's comment, I added pointer declarations for the additional iron bottom flux tracers - this enables values to be saved between time steps. Previously, these values were reset to zero at each timestep. This resulted in fairly sizable discrepancies between `ffetot_btm` and the sum of its constituent parts:
download

With this fix, the two values are much closer together, with discrepancies likely due to the time step issue between the vertical integrals and the bottom diagnostics.

download

yichengt900 (approved):
@jessluo, thanks for the updates! The results look much better now. I agree that the discrepancy in your second plot is more likely due to the time step issue @charliestock mentioned in #109, which will be addressed in a future PR. I approved this PR. @charliestock, feel free to chime in if you have any additional thoughts.

charliestock (commented):
Hi @jessluo and @yichengt900.  This looks good to me.  Thank you both!

Issue #111, opened 2025-01-03
Need to add a namelist parameter to turn on negative oxygen from sulfate reduction


jessluo:
Sulfate reduction is an important redox process that degrades organic carbon in anaerobic sediments. In COBALT, we track sulfate reduction via an oxygen deficit, as negative oxygen. However, this may be confusing to some users as negative oxygen may indicate that there is a bug in the code or the numerics - so some regional configurations are being run with sulfate reduction not being tracked. However, globally, we need to maintain this functionality as sulfate reduction remains a key anaerobic process by which organic carbon is oxidized at the seafloor (potentially up to 30% of organic carbon flux to the seafloor, [Bowles et al. 2014](https://www.science.org/doi/10.1126/science.1249213))

There needs to be an easy way by which the tracking of sulfate reduction can be turned on or off, via a namelist parameter. Right now this is being turned on or off in the code itself: 
https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/135ecc138af87d79a5a748eb02f603ed29edfbb2/generic_tracers/generic_COBALT.F90#L4921-L4923
We need a parameter named something like "track_sulfate_reduction" or "neg_oxygen_on" (ideas, anyone?) that can toggle between the two lines, above.

jessluo:
"track_oxygen_deficit" was suggested as a name.

PR #113, opened 2025-01-18
sulfate-driven remin treatment + some diagnostic cleanup

charliestock:
This pull request cleans up and clarifies the treatment of anaerobic remineralization that occurs after burial, aerobic remineralization and denitrification have been accounted for.  It also implements a first round of diagnostic cleanup.  It was made with input from @jessluo, @rakshit-s and John Dunne.

Material reaching the sediment that was not buried, aerobically remineralized or denitrified in COBALT was previously assumed to be remineralized via other anaerobic pathways, most of which involve Sulfate or iron.  This remainder was only significant when oxygen was completely depleted and was generally a very small part of the global benthic remineralization.

The organic nitrogen remineralized via this pathway was captured in "fnfeso4red_sed" but the stoichiometry and subsequent O2/alk implications were ambiguous.  TOPAZ and prior versions of COBALT allowed this term to generate an oxygen deficit (negative oxygen), presumably associated with hydrogen sulfide, but the details were not clear.  Current CEFI defaults did not consider the alkalinity and o2 impact of sulfate reduction in the sediment (i.e., the creation of a potential latent oxygen demand that would be realized when HS- encountered O2).  This was in part due to ambiguity concerning the processes being invoked, and in part due to an aversion to negative oxygen.  

This pull request explicitly defines the anaerobic process being invoked as 2 step sulfate reduction and defines the stoichiometry of the reaction.   Now that the process being invoked has been clarified and the stoichiometry made explicit, I would recommend including its alkalinity and O2 effects moving forward.  This pull request adds a logical switch “do_fnso4red_sed” which, if set to “.true.” will include these fluxes.  If it is set to “false”, the answers from before this pull request will be reproduced.  Right now, the default is “false” for this reason, but I suggest we move toward “true” as quickly as possible. This addresses issue #111.

The pull request also includes a first round of diagnostic cleanup, aimed in particular at ensuring that all diagnostics are updated from COBALTv2 to COBALTv3.  A number of seldom-used CMIP variables that need updating before CMIP7 and to support CEFI - including the addition of medium phytoplankton to tot_layer_int_poc, and updating a number of “plus_btm” diagnostics needed an updating for “ntot”.  To keep these diagnostics in phase with the code moving forward, the calculation now directly references the bottom fluxes passed to the tridiagonal solver instead of recalculating them.

Finally, as part of this pull request, I began to flag tracer-related diagnostics with “MOVE AFTER TRIDIAGONAL”, with an eye toward addressing issue #109.  I will try and get through the rest in the next week or two so we can put these required changes in motion and address budget woes.


generic_tracers/generic_COBALT.F90
```diff
@@ -4778,8 +4796,16 @@ subroutine generic_COBALT_update_from_source(tracer_list,Temp,Salt,rho_dzt,dzt,h
                   cobalt%zt(i,j,k) / (cobalt%z_denit + cobalt%zt(i,j,k))
 
              ! Calculate the organic matter remineralized via sediment aerobic processes (fnoxic_sed).  This generally
-             ! equals the total flux minus burial and denitrification.  However, if there is insufficient bottom
-             ! oxygen to support this, some is assumed to be remineralized via sulfate reduction.
+             ! equals the total org N flux minus what is lost to burial and what is remineralized via sedimentary 
+             ! denitrification (the second expression in the "min" function below).  However, if there is insufficient
+             ! bottom oxygen, some org N is assumed to be remineralized via sulfate reduction. The max rate of aerobic
+             ! aerobic respiration that can be sustained is based on the specified "bottom thickness" and the bottom
```
yichengt900: duplicated `aerobic`

yichengt900 (commented):
@charliestock, thank you for opening this PR. Overall, the code changes look good and align well with your PR descriptions. I’ll keep it open for now so others can review it and we can discuss it further during the COBALT development meeting.

generic_tracers/generic_COBALT.F90
```diff
@@ -516,6 +516,21 @@ subroutine user_add_params(param_file)
...
+    ! Sulfate reduction: 
+    ! In the sediment, organic material that is not buried, aerobically respired or denitrified
+    ! is assumed to undergo 2-step sulfate reduction: 
+    !   C106H172O38N16 + 59*SO4-- + 75*H+ --> 106*CO2 + 16*NH4+ + 59*HS- + 62*H2O
+    !   59*HS- + 118O2 --> 59*SO4-- + 59*H+
+    ! 
+    ! Yielding the overall reaction:
+    !   C106H172O38N16 + 16*H+ + 118*O2 -> 106*CO2 + 16*NH4+ + 62H2O
+    !
+    ! The effect is an increase of 1 alkalinity equivalent and consumption of 1 mole O2 per mole of
```
rakshit-s: The consumption of O2 would be according to C/N ratio (here Redfield). Here ~6.6 mole O2 will be consumed per mole of N remineralized.

generic_tracers/cobalt_reg_diag.F90
```diff
@@ -1638,8 +1638,8 @@ subroutine cobalt_reg_diagnostics(diag_list,axes,init_time,phyto,zoo,bact,cobalt
     cobalt%id_fntot_btm = register_diag_field(package_name, vardesc_temp%name, axes(1:2),&
          init_time, vardesc_temp%longname,vardesc_temp%units, missing_value = missing_value1)
 
-    vardesc_temp = vardesc("fnfeso4red_sed","Sediment Ndet Fe and SO4 reduction flux",'h','1','s','mol m-2 s-1','f')
-    cobalt%id_fnfeso4red_sed = register_diag_field(package_name, vardesc_temp%name, axes(1:2),&
+    vardesc_temp = vardesc("fnso4red_sed","Sediment Ndet Fe and SO4 reduction flux",'h','1','s','mol m-2 s-1','f')
```
rakshit-s: No `Fe` in comment.

generic_tracers/generic_COBALT.F90
```diff
@@ -516,6 +516,21 @@ subroutine user_add_params(param_file)
...
+    ! N remineralized.  In anaerobic environments this complete reaction does not occur until HS- 
+    ! encounters O2 but, because we do not track S, we allow it to occur locally.  This can result
+    ! in negative O2 values, which should be interpreted as zero oxygen plus additional O2 demand
```
rakshit-s: `... in negative O2 values likely in low bottom water oxygen regions (ODZ or OMZ), ...`

generic_tracers/generic_COBALT.F90
```diff
@@ -4778,8 +4796,16 @@ subroutine generic_COBALT_update_from_source(tracer_list,Temp,Salt,rho_dzt,dzt,h
                   cobalt%zt(i,j,k) / (cobalt%z_denit + cobalt%zt(i,j,k))
 
              ! Calculate the organic matter remineralized via sediment aerobic processes (fnoxic_sed).  This generally
-             ! equals the total flux minus burial and denitrification.  However, if there is insufficient bottom
-             ! oxygen to support this, some is assumed to be remineralized via sulfate reduction.
+             ! equals the total org N flux minus what is lost to burial and what is remineralized via sedimentary 
+             ! denitrification (the second expression in the "min" function below).  However, if there is insufficient
+             ! bottom oxygen, some org N is assumed to be remineralized via sulfate reduction. The max rate of aerobic
+             ! aerobic respiration that can be sustained is based on the specified "bottom thickness" and the bottom
+             ! oxygen concentration ("btm_o2").  The max rate of organic nitrogen remineralization is then:
```
rakshit-s: `The max rate of organic nitrogen remineralization via aerobic pathway is then:`

rakshit-s (commented):
@charliestock Thanks for tagging me in here, I enjoyed going over it, and overall it helped me in familiarizing myself with the code. The sulfate reduction scheme is logical, and a good step forward. I have added a few comments to make the comments more clear. Rest looks good.

yichengt900:
The 15-year (1993-2012) NWA12 simulations show minimal bottom oxygen differences (<0.1 mg/L) in the open ocean with larger variations in shallow regions with `fnso4red_sed=True` vs. `fnso4red_sed=False`.
o2_eval

generic_tracers/generic_COBALT.F90
```diff
@@ -4778,8 +4796,16 @@ subroutine generic_COBALT_update_from_source(tracer_list,Temp,Salt,rho_dzt,dzt,h
                   cobalt%zt(i,j,k) / (cobalt%z_denit + cobalt%zt(i,j,k))
 
              ! Calculate the organic matter remineralized via sediment aerobic processes (fnoxic_sed).  This generally
-             ! equals the total flux minus burial and denitrification.  However, if there is insufficient bottom
-             ! oxygen to support this, some is assumed to be remineralized via sulfate reduction.
+             ! equals the total org N flux minus what is lost to burial and what is remineralized via sedimentary
```
jessluo: @charliestock and I discussed a change to this comment last week, recognizing that what we're simulating here is not the full set of sediment processes, but a subset of processes at the sediment-water interface.

"Calculate the organic matter remineralized via sediment aerobic processes at the sediment-water interface (fnoxic_sed). The thickness of this interface is determined by the "bottom thickness" variable and the total amount of oxygen available within that layer is in the "btm_o2" variable. In reality, oxygen within the sediment becomes depleted very quickly, so the "fnoxic_sed" term also implicitly includes the oxygen demand from local anaerobic processes such as sulfate reduction. However, if there is insufficient bottom oxygen, some organic N is assumed to be remineralized via denitrification and sulfate reduction. The modeled sulfate reduction (“fnso4red_sed”) should thus be viewed as the non-local sulfate reduction that occurs when there is insufficient oxygen within the sediment-water interface as defined by the bottom thickness variable.”

what about something like this?

generic_tracers/generic_COBALT.F90
charliestock: Thanks for catching this @rakshit-s.  I've also been thinking about this description since our last doc & dev meeting.  I think the rationale could be clearer.  What do you and @jessluo think of the following revised comment:

    ! Sulfate reduction/HS- oxidation in the sediment: 
    ! In the sediment, organic material that is not buried or remineralized via denitrification
    ! is assumed to generate oxygen demand at the sediment-water interface through a combination
    ! of direct aerobic remineralization in the sediment (see above) and sulfate reduction followed
    ! by the oxidation of bisulfide (HS-) in the sediment:
    !  
    !   C106H172O38N16 + 59*SO4-- + 75*H+ --> 106*CO2 + 16*NH4+ + 59*HS- + 62*H2O
    !   59*HS- + 118O2 --> 59*SO4-- + 59*H+
    ! 
    ! These two steps yield an overall reaction identical to aerobic remineralization:
    !   C106H172O38N16 + 16*H+ + 118*O2 -> 106*CO2 + 16*NH4+ + 62H2O
    !
    ! Adding 1 alkalinity equivalent per mole of N remineralized. Note that COBALT does
    ! not calculate how much O2 demand is generated by the direct aerobic versus sulfate
    ! reduction/HS- oxidation pathways, just the combined amount. 
    !
    ! If there is no oxygen at the sediment-water interface, organic material is assumed to go
    ! through the first sulfate reduction step in the sediment but not the second, resulting in HS-
    ! release from the sediment. If the logical "do_fnso4red_sed" is set to true, the latent O2 demand
    ! from the HS- release is included in the sediment oxygen sink.  This can result in negative oxygen
    ! concentrations, which should be interpreted as 0 moles O2 kg-1 plus additional O2 demand as HS-.

generic_tracers/generic_COBALT.F90
charliestock: Clarified in the comment above I think.

generic_tracers/generic_COBALT.F90
charliestock: Thanks @yichengt900.  I fixed this and have also edited the comment based on the doc&dev meeting last week.  What do you and @rakshit-s and @jessluo think of:

   ! Calculate the rate of organic matter degradation in the sediment after accounting for burial
   ! and denitrification.  Two pathways are tracked:
   !
   ! fnoxic_sed (moles N m-2 sec-1) accounts for organic material remineralized by processes using oxygen
   ! *within the sediment*. These result in an oxygen demand at the sediment-water interface.  These
   ! include direct aerobic remineralization and sulfate reduction/HS- oxidation (see stoichiometry for
   ! for details).  Note that the partitioning relative importance of these two pathways is not calculated,
   ! just the combined effect. 
   !
   ! fnso4_sed (moles N m-2 sec-1) accounts for organic material that only undergoes sulfate reduction in
   ! the sediment, but not HS- oxidation.  This results in HS- released from the sediment.  The latent O2
   ! demand from the HS- is tracked if "do_fnso4red_sed = true".  The amount of material falling into this 
   ! category is equal to the remainder after all other pathways are accounted for and it generally only
   ! occurs under anoxic conditions in COBALT.
   !
   ! NOTE: The O2 demand from the sediment is calculated by multiplying the organic matter remineralized
   !       by the O2 per N (i.e., fnoxic_sed*o2_2_nh4 or (fnoxic_sed+fnso4red_sed)*o2_2_nh4 if
   !       do_fnso4red_sed is true).
   ! NOTE: fnso4red_sed is not the total sulfate reduction in the sediment, only that which is not paired
   !       with subsequent HS- oxidation.  COBALT does not calculate the total sulfate reduction in seds.
   ! NOTE: The maximum organic remin supported by local O2 is:
   ! btm_o2(moles O2 kg-1)*bottom_thickness(m)*density(kg m-3)* 1/dt(s-1)*molN/molO2 = moles N m-2 s-1
   !
   ! The thickness of the bottom boundary layer (cobalt%bottom_thickness) impacts this upper bound.
   ! Efforts are underway to implement a more dynamic bottom boundary layer scheme.
   !

generic_tracers/generic_COBALT.F90
charliestock: see suggested revision above.

generic_tracers/generic_COBALT.F90
charliestock: Thanks @yichengt900 for the comparison and thanks @jessluo for the comment.  @jessluo and I just discussed her comment in relation to the updated options added to the code above.  We agreed that the updated comments captured the essence of the treatment pretty well.  @rakshit-s, let us know what you think.  If we agree, I will do one last sweep to clean up the grammar and I will update the pull request.   With these refinements and @yichengt900s confirmation of minor impacts in NWA12, I think we'll be able to merge!

generic_tracers/cobalt_reg_diag.F90
charliestock: Thanks, will fix in next commit.

yichengt900 (approved):
Thanks for the updates, @charliestock. Also, thanks for the valuable comments from @jessluo and @rakshit-s. Now, this PR looks great. I fixed a few minor typos. I also checked the `LA-TX hypoxia` zone from the `NWA` run with `do_fnso4red_sed=True`, and it looks good to me. I’ve approved this PR and will merge it later this week if I don't hear any new comments.
hypoxic_area_from_daily

generic_tracers/generic_COBALT.F90
rakshit-s: @charliestock this comment looks nice to me.

jessluo (approved):
Just confirming that I agree with all these changes!

generic_tracers/generic_COBALT.F90
rakshit-s: @charliestock this comment looks good overall, I just have one modification for enhanced clarity: 
Line 4-5 in the above comment: 
! fnoxic_sed (moles N m-2 sec-1) accounts for organic material remineralized by processes which lead to oxygen 
! consumption _within the sediment_.

rakshit-s:
> The 15-year (1993-2012) NWA12 simulations show minimal bottom oxygen differences (<0.1 mg/L) in the open ocean with larger variations in shallow regions with `fnso4red_sed=True` vs. `fnso4red_sed=False`. o2_eval

@yichengt900 Hi thanks for this plot. I was just wondering, if this plot is just flipped (on-off or off-on). Because, if the sulfate reduction is on, I would expect that negative oxygen will lead to more consumption of oxygen in the bottom water, leading to lower oxygen concentration at the bottom water. Currently the figure seems to suggest the other way.

yichengt900:
> > The 15-year (1993-2012) NWA12 simulations show minimal bottom oxygen differences (<0.1 mg/L) in the open ocean with larger variations in shallow regions with `fnso4red_sed=True` vs. `fnso4red_sed=False`. o2_eval
> 
> @yichengt900 Hi thanks for this plot. I was just wondering, if this plot is just flipped (on-off or off-on). Because, if the sulfate reduction is on, I would expect that negative oxygen will lead to more consumption of oxygen in the bottom water, leading to lower oxygen concentration at the bottom water. Currently the figure seems to suggest the other way.

@rakshit-s, thanks for the comment. I had the same concern when I first generated those plots. I just checked the plotting script again and can confirm it is `on-off`. My thought is that in these comparisons, I did not turn off the chl feedback to physics, which could cause physics perturbation and make it difficult for us to isolate the pure impact of the `fnso4red_sed` option. Perhaps I can submit new runs with chl feedback off to ensure there’s no physics perturbation caused by the BGC.

charliestock:
Hi @yichengt900, I suspect we'd need to clean with/without to be sure.  Since we are not advocating this as the default yet, I think we could do this before or after merging.  If we have a new merge by AM on Friday, I will try and build off it by making more progress on the diagnostic cleanup before the next doc&dev.

yichengt900:
Thanks for the input!, @charliestock  I agree—I'll go ahead and merge this PR first so we can keep the cleaning process moving. In the meantime, I'll run a few more tests to dig into the details further.

yichengt900:
I re-ran two quick test cases without `chl` feedback to physics for `fnso4red_sed` set to on and off. Below is the plot showing the bottom oxygen levels along the LA-TX shelf for July 1993. As expected, the major difference occurred in the low oxygen regions, where we observed lower bottom oxygen along most of the LA-TX coast when turn on `fnso4red_sed`. However, a small portion showed a slight increase in bottom oxygen, implying a non-linear impact of the `fnso4red_sed` option.
1993_btm_o2

charliestock:
Thanks @yichengt900 - it looks like a very small impact - considerably smaller than, for example, our previous exploration of the sensitivity to organic matter burial.  

I think the details and impact of this PR are now well documented.  We can wrap things up with a brief discussion at next Tuesday's doc&dev meeting.  Thanks @yichengt900, @jessluo and @rakshit-s for the productive discussion!

PR #114, opened 2025-01-29
Add a Minimum Thickness for the Imbalance Check

theresa-cordero:
After each source/sink update, COBALT checks for an imbalance of elements. This check is done in every grid cell and layer. ESM4.5 developers have found that COBALTv3 requires a much higher tolerance than COBALTv2.

This PR adds an option to specify a minimum thickness for the layer before the check is done. Vanished layers in MOM6 generally do not have trustworthy values, and should not be checked form imbalance. The minimum layer thickness can be set at run time and greater should be greater than the minimum layer thickness set in MOM6.

With the minimum thickness I was able to run ESM4.5 with a imbalance tolerance of 1e-6 for 7 years on C5, while comments suggested 1e-5 was needed to run the model for 5 years. An addition test with a lower tolerance, 1e-9, is on going.


charliestock:
Thanks @theresa-morrison, this is a very promising result!  I had two quick questions:

1) Do you have any thoughts on the best default value for min_thickness?  I see you chose 0.001m, which seems reasonable, but had me wondering why...

2) Do you need the local variable "min_thick" or could we just use cobalt%min_thickness in the if statement?

Thanks!
Charlie

yichengt900:
@charliestock, thanks for the comments. Not directly related to this PR, but I do have a minor suggestion regarding the [conversion](https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/e4e198909da8b97f4304dda14afe43c290e873ad/generic_tracers/generic_COBALT.F90#L5598) `86400.0/dt*1.03e6` used in the imbalance calculation. Would it be beneficial to add a clear comment explaining this conversion in the future? This could help improve clarity and ensure a better understanding of the imbalance unit for others working with the code.

charliestock:
Good point @yichengt900.  I will add this before we finalize.

On a related note - could you provide the github instructions necessary for me to add a commit with the comment requested to Theresa's pull request?  Alternatively, I could wait and add one once we merge this one.  

So we all know, the conversion takes the imbalance over a time step in moles kg-1 and converts it to ~ mmoles m-3 day-1.  This means the imbalance check is not sensitive to the time step and has units people are used to thinking about: e.g. typical plankton concentrations are ~0.1-1 mmoles N m-3 day-1, so an imbalance of order 1 would be very large whereas 1e-9 is very small.

Other tracers can have values that are quite a bit larger or smaller than N but values ~1e-7 - 1e-9 seemed sufficiently small in all cases.  1e-5 starts to get a bit too large for tracers like iron that have small concentrations already.

theresa-cordero:
>     1. Do you have any thoughts on the best default value for min_thickness?  I see you chose 0.001m, which seems reasonable, but had me wondering why...

0.001m is the default minimum thickness in MOM6 when regridding the layers. 
 
>     2. Do you need the local variable "min_thick" or could we just use cobalt%min_thickness in the if statement?

We could put `cobalt%min_thickness` in the if statement.

I can add another commit with that change and a comment about the conversion. 

The ESM4.5 run with 1e-9 tolerance is in year 6!

yichengt900:
@theresa-morrison and @charliestock, thanks for all the input. I believe you should be able to commit directly to [theresa-morrison:add_min_thickness_check_to_src_checks](https://github.com/theresa-morrison/ocean_BGC/tree/add_min_thickness_check_to_src_checks) since @theresa-morrison has enabled "Allow edits from maintainers." However, to keep things straightforward, I suggest we leave a suggested comment here so she can implement it in her next commit.

charliestock:
Sounds good @yichengt900.  Theresa do you have everything you need for a comment based on my last e-mail?  We should also make sure the units are indicated on the imbalance tolerance (effectively mmoles m-3 day-1)

charliestock:
Sounds good @yichengt900.  Theresa do you have everything you need for a comment based on my last e-mail?  We should also make sure the units are indicated on the imbalance tolerance (effectively mmoles m-3 day-1)

theresa-cordero:
Unfortunately the imbalance is set in the input namelist, so we can't specify the units there. Let me know if the comment based on you comments above should be revised.

yichengt900:
@theresa-morrison, the changes look good to me. @charliestock, @theresa-morrison made a good point that `imbalance_tolerance` is currently set in the namelist. Since COBALT is the only module using this specific namelist, should we consider moving it to our MOM6-like runtime parameters eventually? No need to do it in this PR, but it's something we can think about to make COBALT runtime parameter setup more consistent.

charliestock:
Looks good to me @theresa-morrison and @yichengt900.  I agree that we should eventually move the imbalance to a runtime parameter, but lets hold off on that for now to take it easy on Eric/ESM4.5.  Perhaps we should see how Theresa's test cases look tomorrow and then merge?

theresa-cordero:
My short test of 7 years 1e-9 has finished successfully!

yichengt900:
Thanks for the update, @theresa-morrison! Everything looks good to me. I will approve this PR and plan to merge it by the end of the week. After that, we will implement major tag updates on CEFI-regional-mom6 to reflect recent code updates.

yichengt900 (approved):

PR #117, opened 2025-02-21
First attempt to improve global budget diagnostics and aggregated btm fluxes

yichengt900:
This draft PR addresses #109 and introduces several changes:

1. Added an optional argument, `post_vertdiff`, to cobalt_send_diagnostics. When set to `.true.`, it triggers the calculation of tracer water column integrals.
2. Updated `generic_COBALT_update_from_bottom` to call `cobalt_send_diagnostics` with `post_vertdiff=.true.`, ensuring the water column integral is calculated using the updated tracer fields.
3. Recalculated aggregated bottom fluxes in `generic_COBALT_update_from_bottom`. In this draft PR, only major bottom fluxes—such as `fntot_btm`, `fptot_btm`, `ffetot_btm`, and `fsitot_btm`—are recalculated. We can extend this to other benthic flux calculations if we determine that this  code refactoring approach is appropriate.

@charliestock, feel free to take a look and let me know your thoughts.

**Update**
~~It looks like this PR did not pass RT. I will need to investigate further to determine the issue. For now, please disregard this draft PR.~~ Ok, after consolidating the `post_vertdiff` logic, this PR passed RT.


charliestock:
Thanks @yichengt900, the logic looks sound to me.  Would it be possible for @jessluo to pull from this branch to run some tests to see if the budget diagnostics close more tightly after this PR?  How would you modify the xml to grab this rather than the standard code?

I would be OK with merging, as this would only impact the water column integral and the bottom flux diagnostics slightly (and make them more correct), but maybe we should test first?

yichengt900:
It should be straightforward to modify your XML to check out the branch. For instance, in the global 025 XML, add the following lines after[ line 104](https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/regional-mom6-xml/blob/ce3b3187e494ddfe177f0256a4aa5efebf5fc97a/global/OM4_025_COBALT_4P_jra55_const_c6.xml#L105):
```
pushd mom6
(cd src/ocean_BGC &&  git checkout feature/budget)
popd
```
Then, redo the `fremake` (do not forget to remove your old source code or perform a force checkout) and re-run the build script.

jessluo:
I ran a test with this PR for 20 years in ocean-only mode with a global 1/2 degree ocean configuration. It definitely appears that the budgets are closing much more tightly. (The only one I'm having trouble with is alkalinity, which I'm still troubleshooting.) Here are the associated plots for the various budgets, with associated flux imbalance plots:

### Nitrogen
download
download
The flux imbalance plots show fluctuations about 1 order of magnitude smaller than the previous run. (This is fairly consistent across all the budget calculations.) The flux imbalance is calculated by taking the difference between the fluxes as calculated by the year-to-year change in the total N inventory and the fluxes as calculated by the sum of all the N fluxes (runoff + deposition + nitrogen fixation + icebergs - denitrification - anammox - burial).

### Carbon
download
download

### Phosphorus
download
download

### Oxygen
download
download

### Iron
download
download

However, it seems like the inventory imbalance plots are generally showing positive values for all the budgets, except oxygen. The inventory imbalance is calculated by taking the difference between the annual vertical integral (for N, for example) and the inventory as calculated by the sum of the fluxes plus the starting N inventory.
#### Nitrogen
download
#### Carbon
download
#### Phosphorus
download
#### Oxygen
download
#### Iron
download

Keep in mind, however, that these imbalances are about 1-2 orders of magnitude smaller than prior to this PR. So I would not initially think that these inventory imbalances are of any concern. However, I'm planning to run this run out further, to at least 50-60 years, to see if these imbalances keep growing. 

The alkalinity budget is also forthcoming, once I finish troubleshooting that budget code.

charliestock:
Hi @jessluo, Thank you for this.  They look much improved.  Let's tackle the alkalinity a bit more when we catch up tomorrow and discuss the accumulation.   Have a good night.

jessluo:
In discussion with @charliestock, there does not appear to be anything obviously wrong with the way my [alkalinity budget codes](https://github.com/jessluo/COBALT_diagnostics/blob/main/notebooks_esm45/budget_alkalinity.ipynb) are written up. It appears that perhaps, the alkalinity issue may (??) be just due to a small imbalance in an overall flat alkalinity trend. Below are the alkalinity plots, again from the 1/2 degree forced ocean model run, which has been run out for 30 years:

download

download

download

jessluo:
Additionally, I have done some further runs with the current candidate model of ESM4.5 (1/4 degree ocean) that has run out for 40 years. Aside from one minor issue, where the iceberg iron fluxes are not being written in correctly, the overall elemental budgets all look quite good.

### Nitrogen
download
download
download


### Carbon
download
download
download


### Phosphorus
download
download
download


### Oxygen
download
download
download


### Iron
For the Iron budgets, the iceberg iron is currently not written in correctly in ESM4.5, which is causing an imbalance. However, since the iron budget closes quite well in the ocean-only run, I think we can be confident that once the iceberg iron gets written in correctly then this budget will also close.
download
download
download

### Alkalinity
Again, there appears to be an imbalance in the alkalinity. However, the budget imbalance seems to alternate between positive and negative (2 positive data points) so there is a chance that this may not indicate that we have a systematic alkalinity sink missing in the budget codes. However, it is still important to keep an eye on this budget to see how these imbalances play out.
download
download
download

charliestock:
Hello @jessluo, @yichengt900, @feida6996 and @nikizadehgfdl,

Great discussion at today's doc&dev meeting.  Thank you @yichengt900 for putting together this PR and thanks @jessluo for putting these plots together! 

I was staring at the code a bit while my daughter was swimming laps and I think I may have found a potential culprit that could solve this puzzle.  This pull request set all of the bottom flux diagnostics so that they corresponded to the flux to the bottom during the time steps between the instantaneous water column inventory snap shots.  This is correct for understanding the flux into the sediment over the time period.  The flux out of the sediment, however, is a lag 1 response to the flux at the previous time step.  Thus, to fully understand the budgets we need to save both the fluxes *into* and *out of* the sediment.

Take b_alk, for example:

https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/cf1d8412bec2886af85ff0e92430ddc0d377904d/generic_tracers/generic_COBALT.F90#L4977-L4978

It includes cobalt%f_cadet_arag_btf(i,j,1) from the last time step, so we are offset by 1 relative to what we now save.  If, however, we redefine:

fcased_redis = fcased_redis +  cobalt%f_cadet_arag_btf(i,j,1)

(i.e., the it is now the combined redissolution of calcite and aragonite), we could capture the right fluxes in fcased_redis.  This would also improve b_dic:

https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/cf1d8412bec2886af85ff0e92430ddc0d377904d/generic_tracers/generic_COBALT.F90#L4969-L4970

At the end of the day, this issue is injecting imprecision into nearly all of our budget diagnostics, so I think we may just need to bite the bullet and save the bottom fluxes as their own diagnostics to ensure that everything is synced up:

https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/cf1d8412bec2886af85ff0e92430ddc0d377904d/generic_tracers/generic_COBALT.F90#L4969-L4985

Thoughts?  I'm hopeful that this may get us to the precision we are after!

yichengt900:
@charliestock, good point. I’ve updated this draft PR, and now we have extra diagnostics for `b_alk`, `b_dic`, `b_fed`, `b_no3`, `b_nh4`, `b_o2`, `b_po4`, and `b_sio4`.

You can follow the example below to add these diagnostics to your `diag_table`:
```
 "generic_cobalt","b_alk",     "b_alk",    "ocean_cobalt_tracers_instant","all",".false.","none",2
 "generic_cobalt","b_dic",     "b_dic",    "ocean_cobalt_tracers_instant","all",".false.","none",2
 "generic_cobalt","b_fed",     "b_fed",    "ocean_cobalt_tracers_instant","all",".false.","none",2
 "generic_cobalt","b_nh4",     "b_nh4",    "ocean_cobalt_tracers_instant","all",".false.","none",2
 "generic_cobalt","b_no3",     "b_no3",    "ocean_cobalt_tracers_instant","all",".false.","none",2
 "generic_cobalt","b_o2",     "b_o2",    "ocean_cobalt_tracers_instant","all",".false.","none",2
 "generic_cobalt","b_po4",     "b_po4",    "ocean_cobalt_tracers_instant","all",".false.","none",2
 "generic_cobalt","b_sio4",     "b_sio4",    "ocean_cobalt_tracers_instant","all",".false.","none",2
```

These should give us the exact benthic fluxes back into the ocean that are used by `g_tracer_vertdiff_G`. Let me know if anything is unclear. I will conduct some tests later, but I’d appreciate it if any of you could give it a try and see if it improves the budget imbalance calculation.

yichengt900:
Ok, so it looks like we can achieve a better `alkalinity `budget imbalance with the new `b_alk` diagnostic:

global_alk_inventory_nosink
global_alk_budget_imbalance_nosink



The plot above shows preliminary results from an ocean-only 0.5° run using the new budget diagnostics. In this specific test case, I turned off phytoplankton sinking (only detritus sinks in this run) and kept atmospheric CO₂ at pre-industrial levels.

I also have another ongoing run with sinking phytoplankton, and so far, the alkalinity budget  imbalance also looks ok to me:
global_alk_inventory_sink
global_alk_budget_imbalance_sink




Overall, the results are promising, but I want to wait for longer runs and have others conduct more comprehensive checks before drawing conclusions.

charliestock:
Looks excellent @yichengt900.  Thanks for running these tests.  Let's confirm the N, P, O2 and iron etc. next week.  Based on @jessluo's prior tests and these results, I am optimistic that we are in good shape.

charliestock:
I just wanted to put a placeholder here to discuss the alignment of the surface flux diagnostics for the budget calculations as well.  We don't currently have the surface flux calculation in the flow diagram, but we should probably ensure that the surface fluxes we save in cobalt_send_diags are those that are applied for the current time step and not the prior.  @theresa-cordero, @yichengt900, @jessluo

jessluo:
Closing the loop on the updated PR with the explicit `b_` bottom flux terms:

After running an ocean-only case for ~40 years, it looks like the budgets close more tightly than before. However, the difference that having the explicit `b_` terms give us is much smaller than anticipated (approximately 4 orders of magnitude smaller than the flux imbalances, at least for carbon), and it also does not remove the directional trend in the inventory imbalances that we were seeing in some of the budgets.

For example, for **CARBON**:
download
The global carbon inventory as outputted from the vertical integral terms vs. calculated using the carbon fluxes still match very well. 
download
The flux imbalances have positives and negatives, indicating no missing source or sink. 
download
Looking at the differences in the carbon fluxes with and without the `b_dic` term, it appears that the differences are on the order of 1e9, about 4 orders of magnitude smaller than the flux imbalances.
download
And the inventory imbalance (in g C) unfortunately still shows a directional trend that grows over time.


The **ALKALINITY** budget closes fairly well:
download
download
With positives and negatives in the alkalinity budget imbalance.
download
Though, the inventory imbalance is still overall negative.
However, for the sake of completeness, I would like to note that previously, the alkalinity budget did not close because the impact of anammox on alkalinity was directionally reverse (anammox was supposed to remove alkalinity in the budget but instead, I had it as adding alkalinity). When that was fixed, then the budget as computed without the `b_alk` term also closed.


**NITROGEN** also closes, but shows an overall positive inventory imbalance that is roughly stable after the first ~5-7 years:
download
download
download


**PHOSPHORUS** also shows a growing inventory imbalance:
download
download
download


Whereas **OXYGEN** shows an imbalance that gets more negative over time:
download
download
download


Similarly with **IRON**:
download
download
download

Incidentally, when looking at the iron budgets as calculated with and without the `b_fed` term, the difference was still non-zero:
download
But this value is summed up over the whole globe over a year, so the actual difference in g Fe / m2 / s is minuscule and likely close to a roundoff level of error.

yichengt900:
@jessluo, thanks for all the new plots for the budget check! When you get a chance, could you update the scripts you used for the budget check? I think the current script you shared is probably out of date. For example, if I use your [Alk script,](https://github.com/jessluo/COBALT_diagnostics/blob/main/notebooks_esm45/budget_alkalinity.ipynb) I get a value of 0 for the Annamox influence on Alk (derived from `wc_vert_int_jnamx`). It would be nice to use the same scripts to ensure the plots are generated using exactly the same diagnostics.
alk_process

jessluo:
> @jessluo, thanks for all the new plots for the budget check! When you get a chance, could you update the scripts you used for the budget check? I think the current script you shared is probably out of date. For example, if I use your [Alk script,](https://github.com/jessluo/COBALT_diagnostics/blob/main/notebooks_esm45/budget_alkalinity.ipynb) I get a value of 0 for the Annamox influence on Alk (derived from `wc_vert_int_jnamx`). It would be nice to use the same scripts to ensure the plots are generated using exactly the same diagnostics. alk_process

Yes, absolutely. I will update it right away. With anammox, though, it is turned off by default in COBALTv3, and to turn it on you will need to add `#override gamma_nh4amx = 0.07` to your `COBALT_override` file.

charliestock (approved):
Hello @yichengt900, as discussed at today's doc&dev meeting, I think we can move ahead with merging this and building off the structure it provides for the other diagnostics.  Thank you!

Issue #120, opened 2025-02-27
Use "o2_2_nfix" instead of "o2_2_nh4" for oxygen production during N2 fixation?


yangyzhao5:
The coefficient "o2_2_nfix" is defined with a value of 130.0/16.0, but currently the calculation of total net oxygen production in water column includes oxygen production during nitrogen fixation using a conversion of "o2_2_nh4" (118.0/16.0) for "juptake_n2". This might affect oxygen concentrations in the upper ocean.

```
do k = 1, nk ; do j =jsc, jec ; do i = isc, iec  !{
   cobalt%jo2(i,j,k) = (cobalt%o2_2_no3 * (phyto(DIAZO)%juptake_no3(i,j,k) +   &
        phyto(LARGE)%juptake_no3(i,j,k) + phyto(MEDIUM)%juptake_no3(i,j,k) + &
        phyto(SMALL)%juptake_no3(i,j,k)) + cobalt%o2_2_nh4 *    	&
        (phyto(DIAZO)%juptake_nh4(i,j,k) + phyto(LARGE)%juptake_nh4(i,j,k) +  	&
        phyto(MEDIUM)%juptake_nh4(i,j,k) + phyto(SMALL)%juptake_nh4(i,j,k) + &
        phyto(DIAZO)%juptake_n2(i,j,k))) * grid_tmask(i,j,k)
   cobalt%jo2(i,j,k) = cobalt%jo2(i,j,k) - cobalt%jo2resp_wc(i,j,k)
   cobalt%p_o2(i,j,k,tau) = cobalt%p_o2(i,j,k,tau) + cobalt%jo2(i,j,k) * dt * grid_tmask(i,j,k)
enddo; enddo ; enddo  !} i,j,k
```

charliestock:
You appear to be right @yangyzhao5, but let me check with John Dunne on this , as this treatment goes back to TOPAZ.

Issue #119, opened 2025-02-27
Negative oxygen values in the oxygen minimum zones


yangyzhao5:
We ran our simulations based on MOM6-COBALTv3 with modifications from 1958-1969:
1) Added a new tracer tsldon (terrestrial semi-labile dissolved organic nitrogen) that can be degraded to marine sldon and participate in Iron ligand formation;
2) Switched to a wind-wave-bubble (D25) scheme for air-sea gas exchange (CO2 and O2);
3) Turned on the photoacclimation effect in MOM_input, using a density difference to find the photoacclimation mixed layer depth.

We found an extensive area of negative oxygen values at shallow depths (e.g., 250 m) of oxygen minimum zones (OMZs, i.e., the Arabian Sea and Eastern Tropical Pacific). The oxygen concentrations are declining over 1958-1969 with an increasing area of negative oxygen values (black and dark red lines indicate oxygen concentrations of 0 and 20 umol/kg, attached figure below at 250 m and 800 m in August of 1960 and 1969).

![Image](https://github.com/user-attachments/assets/2b218c20-a147-42b6-8a94-5ad62be15220)

These negative oxygen values present at shallower depths (\~250 m) than that in COBALTv2 (\~800m) - that we've used previously with the D25 scheme.

The oxygen demand for oxidation of HS- from sulfate reduction has been excluded/commented if I understand the codes right.

```
      	! uncomment if you want to include sulfate reduction
      	!cobalt%b_o2(i,j)  = cobalt%o2_2_nh4 * (cobalt%fnoxic_sed(i,j) + cobalt%fnfeso4red_sed(i,j))
      	cobalt%b_o2(i,j)  = cobalt%o2_2_nh4 * cobalt%fnoxic_sed(i,j)
      	cobalt%b_po4(i,j) = - cobalt%fptot_btm(i,j) + cobalt%fp_burial(i,j)
      	cobalt%b_sio4(i,j)= - cobalt%fsitot_btm(i,j)
```

I checked the growth rate of phytoplankton (phyto(n)%mu) where oxygen values are negative. These growth rates are negative. I wonder if negative oxygen values in the OMZs are caused by the oxygen consumption by respiration of phytoplankton, zooplankton and higher prey when oxygen is depleted (e.g., O2 < 10 nmol/kg).

PR #124, opened 2025-03-03
extra checks to prevent o2 consumption under anoxic conditions + o2_2_nfix fix

charliestock:
Issue #119 showed areas of slightly negative oxygen in anoxic zones.  This was shown to be beyond what was expected from the O2 demand associated with sulfate reduction in the sediments.

Inspection of the code suggested that it was possible to have low levels of net plankton respiration in anoxic regions when growth rates were negative.  This was not a significant issue in COBALTv2 because phytoplankton did not sink.  Now that they do, however, the issue is more prominent.

This pull request addresses the issue by imposing additional limiters on aerobic respiration by plankton in low O2 waters.  For phytoplankton, net respiration only proceeds when o2 > o2_min.  If o2 < o2_min, phytoplankton exude labile organic carbon rather than respire it.  These changes can be found distributed between lines 3544-3600 of generic_COBALT.F90.  The exudation calculations between lines 4230-4237 also required an update to add new values to any amounts that were already added (i.e., they are now cumulative).

An oxygen limiter was already included on the zooplankton ingestion term, so this should not be contributing to issue #119.  This pull request adds some comments between lines 4418-4429 to clarify the logic.  We should run some longer simulations to confirm there are no issues.  An o2 limiter on higher predator respiration, however, was added between lines 4461-4470.  All material ingested by "higher predators" results in detritus under anoxic conditions.

Finally, this pull request corrects replaces the use of o2_2_nh4 for oxygen creation associated with N fixation to o2_2_nfix - changing it from 118/16 to 130/16.  This addresses issue #120.  

Thank you @yangyzhao5 for raising these issues.  It would be great if you could run some tests alongside those that we have started to confirm that this PR will fully address issue #119.


charliestock:
I should have mentioned that this PR is answer changing, so it may not pass all checks!

generic_tracers/generic_COBALT.F90
```diff
@@ -4417,7 +4443,7 @@ subroutine generic_COBALT_update_from_source(tracer_list,Temp,Salt,rho_dzt,dzt,h
           ! Add respiration-associated excretion to the cumulative production of inorganic nutrients
           cobalt%jprod_nh4(i,j,k) = cobalt%jprod_nh4(i,j,k) + zoo(m)%jprod_nh4(i,j,k)
           cobalt%jprod_po4(i,j,k) = cobalt%jprod_po4(i,j,k) + zoo(m)%jprod_po4(i,j,k)
-          ! Zooplankton respiration uses oxygen 
+          ! Zooplankton respiration uses oxygen (note zoo%jprod_nh4=0 when o2<02_min from zoo%o2_lim)
```
yichengt900: Great note! Just a small suggestion: the variable name for the oxygen limiter in zooplankton is `zoo%o2lim` in current code. I'm not trying to be picky, but using the exact variable name might make it easier for users and developers to search for the keyword.

yichengt900:
@charliestock, thanks for your quick response to this issue and for coming up with a solution. The overall code changes make sense to me. I’ll run global high-resolution simulations to evaluate these changes. As we discussed, I understand that COBALTv3 is still under scientific evaluation, and any aspects of the code may undergo significant changes (both in infrastructure and scientific development), which could lead to baseline changes. I’m okay with that for now, but eventually, I think we should adopt a similar approach used by the MOM6 group, such as introducing a `bug` flag or an `ANSWER_CHANGE` flag. This would help us easily track changes and ensure we can reproduce the old results when needed.

**Update**:
Hmm, looks like I encountered an imbalance error for nitrogen with the new changes in both the OM4 0.5deg and 0.125deg tests. I’ll need to take a further look...

yangyzhao5:
@charliestock , thanks for your quick response to this issue. The modified codes make sense to me. I will run a test simulation to see whether it works well. Only one small question: should o2_min_nit = 0.01 umol/kg be used for the oxygen threshold of anoxic conditions since o2_min = 0.8 umol/kg is the oxygen threshold for aerobic respiration while below this threshold denitrification can proceed?

charliestock:
Hi folks, this has been tripping the conservation checks in larger simulations.  I think I found the reason why.  I will put in another commit tonight that should address it! @yangyzhao5 @yichengt900

charliestock:
Hello @yangyzhao5 and @yichengt900.  I believe I was able to address the conservation issue arising from the initial pull request.  The issue was that net negative growth is already accounted for in the phytoplankton time stepping equation.  When the negative growth was added to exudation in near 0 O2 environments, this resulted in a double counting.  This commit circumvents that by adding the phytoplankton mortality to cobalt%jprod_sldon, attributing it to cell death, and keeping the exudation as it was before.  This prevents the double counting.

I have tested this for several years in a global run without problems, but please run some tests on your ends as well.  It seems that the mass conservation check has done its job and ensured the PR did not introduce a conservation issue.

@yangyzhao5, with regard to o2_min, we are waiting on a new nitrification formulation from Weiyi Tang.   Fabian Paulot implemented a separate parameter for nitrification some time, but I was never completely convinced we needed one.  I would rather link to o2_min and modify this value as needed.  I will pester Fabian and Weiyi on their formulation update and whether we can simplify o2_min.

Let me know what you think.

yichengt900:
@charliestock, thanks for the patch! It looks like it passed the imbalance check. I’ll conduct a few long-term tests and get back to you.

yangyzhao5:
@charliestock Thanks for the updates and the model is running successfully now. I will get back to you when I gets several years of model outputs.

Regarding the o2_min, observations revealed that nitrification can proceed at a nanomolar oxygen concentration. Also, aerobic respiration can also proceed at oxygen levels lower than 0.8 umol/kg. But I agree with you to keep it as of now to use o2_min in this case.

yichengt900:
Ok, here are the preliminary results from the 10-year, 0.25-degree run. It looks like the negative oxygen at intermediate water has been mitigated after the new changes. I’ll keep the run going for a while to make sure everything is good
image

yangyzhao5:
@charliestock @yichengt900 Thanks for the updates. I also checked the oxygen after running the OM4p5 for 5 years and there are almost no negative oxygen values except only several cells along the eastern tropical Pacific coasts (attached below). I will continue running the model some years more to track the oxygen distributions. Thank you!

Screenshot from 2025-03-05 10-08-05

yichengt900:
Ok, I have completed a 27-year run, and the oxygen at the intermediate water looks fine to me:

image

If I don’t hear any further concerns about this PR, I will update the baseline answers accordingly and then merge it.

yichengt900 (approved):

charliestock:
Thanks for the thorough evaluations of this PR @yichengt900 and @yangyzhao5, and thank you for raising the issue @yangyzhao5.  I'm glad the fixes that we implemented seem to have addressed the issue.

Issue #125, opened 2025-03-04
Duplicated but unreachable budget diagnostics


andrew-c-ross:
1. generic_tracer_utils has a subroutine [`g_tracer_diag`](https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/cf1d8412bec2886af85ff0e92430ddc0d377904d/generic_tracers/generic_tracer_utils.F90#L3035) that, as far as I can tell, is unreachable. It is called within [`generic_tracer_diag`](https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/cf1d8412bec2886af85ff0e92430ddc0d377904d/generic_tracers/generic_tracer.F90#L440), but I don't think anything ever calls `generic_tracer_diag`. 

2. generic_tracer_utils has a subroutine [`g_tracer_register_diag`](https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/cf1d8412bec2886af85ff0e92430ddc0d377904d/generic_tracers/generic_tracer_utils.F90#L1200) which registers at least one diagnostic (_tendency) that has the same name as a diagnostic on the MOM6 side. The _tendency diagnostic is written to in the `g_tracer_diag` routine that never gets called. Separately, MOM6 does write to its own _tendency diagnostic in [`post_tracer_diagnostics_at_sync`](https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/MOM6/blob/cf4f6d8c80b94ab563e30e73b4c89a0e038ed3cb/src/tracer/MOM_tracer_registry.F90#L661). This means there appears to be _tendency diagnostics available for both the "ocean_model" and "generic_cobalt" components, but if you choose the "generic_cobalt" component in the diag table the output will be all zeroes.

PR #126, opened 2025-03-06
Remove unused diags (tendency and taup1) and subroutines

andrew-c-ross:
Cleans up issue #125 by removing the unreachable subroutine `g_tracer_diag` and its parent `generic_tracer_diag`, plus the two unwritten diagnostics they register (`_tendency` and `_taup1`). Note this does not remove `diag_id_vdiff` because that does get written to in `g_tracer_send_diag`.


yichengt900 (approved):
@andrew-c-ross, thanks for cleaning up unused diagnostics and subroutines. The code changes look good by eyes. Since `g_tracer_diag` only covers three diagnostics—`taup1`, `tendency`, and `vdiffuse_impl`—and I don’t think we use `taup1`, plus we can get `tendency` from MOM6 and `vdiffuse_impl` from g_tracer_send_diag, it should be safe to remove those unsed subroutines. Approved.

PR #128, opened 2025-03-13
Initial generic_CBED  module for benthos calculations

nikizadehgfdl:
- Refactorization of generic_COBALT.F90 to make it ready to use generic_CBED module for benthos canculations.
- This commit merges in the CBED refactorization made by @laurenchilutti to CEFI repo/branch of ocean_BGC
- Based on https://github.com/laurenchilutti/ocean_BGC/tree/CBED_step3
- Because of extensive changes and reorg in dev/cefi relative to the original ocean_BGC I could not do the merge automatically with git and did it manually.
- Currently do_CBED=.true. and .false. are bitwise identical.
- No answer changes in ocean-ice-COBALT test case


nikizadehgfdl:
@laurenchilutti  please review to make sure all your work got in this PR. Because of extensive differences between noaa/ocean_BGC and CEFI/ocean_BGC repos I was not able to do git merge and had to do it manually based on your branch.

@rakshit-s , @jessluo  please review to make sure the two calls are at the right place and  with the right content:
generic_CBED_sediments_update_from_bottom
generic_CBED_sediments_update_from_source

jessluo (commented):
Following our discussion today with @nikizadehgfdl @rakshit-s @laurenchilutti @charliestock, we decided the following:
1) the main call to CBED will be in the `generic_COBALT_update_from_source` subroutine, to be placed before the Source/Sink calculations in section 6 (e.g., `generic_CBED_sediments_update_from_source`). This is equivalent to placing this CBED call at the bottom of `generic_tracers_update_from_bottom` but with the added benefit of not having to modify the restart files. We thought it would be cleaner to do it this way. 
2) `generic_CBED_sediments_update_from_source` will not be modifying the COBALT tracers (in `cobalt%` and `phytoplankton%` types) but will be passing the bottom flux values (as `b_{tracer}`) back to COBALT as a bottom boundary condition. COBALT types `cobalt` and `phytoplankton` will still be passed to CBED with `intent(inout)` but will just not be modified in CBED.
3) We do not need a generic_CBED_sediments_update_from_bottom subroutine, and the if statements in lines 2557-2559 of `generic_COBALT.F90` can be removed.
4) We will need a `CBED_init` subroutine that will read from the restarts or initial conditions and registers the diagnostics, as well as a `CBED_end` which writes the restart files, in addition to the `CBED_update_from_source` subroutine. For the diagnostics, we can use the existing diag manager to register and write diagnostics.
5) The CBED vertical grid can be defined at each `CBED_update_from_source` call because it will be a simple array, and it will be easier to do it within the subroutine than outside.
6) On the science side, since CBED does not include any calculations for the inorganic carbon cycle, we will only use CBED for the organic carbon transformations, and use the existing Dunne et al. 2012 calcium carbonate burial/dissolution meta model. Alkalinity changes from the organic matter degradation will be calculated first, then passed to COBALT to be updated with calcium carbonate redissolution from sediments.

Hopefully I captured the main points. Please comment if there is anything I missed!

charliestock:
This is consistent with my recollection @jessluo, thanks for synthesizing.  

I would just further note that, under the proposed approach, the CBED state derived at time step "t" are the conditions underlying the bottom water column tracer boundary condition (i.e., b_{tracer} terms) at time step "t".  This calculation, however, is based on the sinking fluxes from from the water column and water column state (i.e., bottom O2) from time step "t-1".

In effect, CBED is being updated first based on prior water column fluxes and conditions.  The bottom boundary condition it provides is then allowing us to update COBALT.   I think this sequential solution is OK and unavoidable.

best,
Charlie

generic_tracers/generic_COBALT.F90
```diff
@@ -2543,17 +2545,22 @@ end subroutine generic_COBALT_update_from_coupler
   !  </IN>
   !
   ! </SUBROUTINE>
-  subroutine generic_COBALT_update_from_bottom(tracer_list, dt, tau, model_time)
+  subroutine generic_COBALT_update_from_bottom(tracer_list, dt, tau, model_time,Temp, rho_dzt, dzt, ilb, jlb)
```
nikizadehgfdl: @yichengt900 I did not change this subroutine. The changes for these subroutine are in dev/cefi. I don't know why they are appearing in this PR!

generic_tracers/generic_COBALT.F90
```diff
@@ -4701,30 +4758,21 @@ subroutine generic_COBALT_update_from_source(tracer_list,Temp,Salt,rho_dzt,dzt,h
 ! 5: Sediment, coastal and ice dynamics
 !-------------------------------------------------------------------------------------------------
 !
+    ! Local variables used to determine the layers falling within the bottom thickness
```
nikizadehgfdl, line 4764: Must be allocated outside if(do_CBED)

generic_tracers/generic_COBALT.F90
```diff
@@ -4701,30 +4758,21 @@ subroutine generic_COBALT_update_from_source(tracer_list,Temp,Salt,rho_dzt,dzt,h
...
+    allocate(rho_dzt_bot(isc:iec,jsc:jec))
+    allocate(k_bot(isc:iec,jsc:jec))
 
-    ! Nutrient inputs associated with icebergs/frozen runoff.  This is currently entered as a surface flux.  The
```
nikizadehgfdl, line 4760: Move further down closer to where it's used for clarity.

Issue #122, opened 2025-02-27
repeated calculation of cobalt%btm_o2


nikizadehgfdl:
@jessluo @charliestock
There is a repeated calculation of some  cobalt%btm_ values (e.g., cobalt%btm_o2) in two different places in generic_COBALT.F90.
Once circa [line 4742](https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/dev/cefi/generic_tracers/generic_COBALT.F90#L4742)
 and then other circa [line 6434](https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/dev/cefi/generic_tracers/generic_COBALT.F90#L6434C10-L6434C23)

I am wondering which is the correct place to calculate these quantities (for the benefit of consolidating them for the future CBED module). We probably don't need both calculations, do we?

PR #129, opened 2025-03-14
clean up duplicate btm calculation

yichengt900:
This PR addresses issue #122. We cleaned up some of the duplicate btm calculations. This PR passed the 1D regression test, and the btm diagnostics are identical to those from the current `dev/cefi` branch

### DIFF
--- a/generic_tracers/generic_COBALT.F90
+++ b/generic_tracers/generic_COBALT.F90
@@ -4743,6 +4743,8 @@ contains
           rho_dzt_bot(i,j) = 0.0
           cobalt%btm_o2(i,j) = 0.0
           cobalt%btm_no3(i,j) = 0.0
+          cobalt%btm_co3_sol_calc(i,j) = 0.0
+          cobalt%btm_co3_ion(i,j) = 0.0          
           cobalt%btm_omega_calc(i,j) = 0.0
           k_bot(i,j) = 0
           ! Note that grid_kmt is always the total number of layers in MOM6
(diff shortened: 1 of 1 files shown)


charliestock:
Hi @yichengt900 and @nikizadehgfdl.  Sorry for the delay in addressing this issue.  I actually think we may need both.  The first call calculates these quantities for the purpose of influencing the sediment processes during the current time step.  The second call is for the purpose of saving the diagnostic after quantities have been updated.  We should move the second call to after the tridiagonal since it is tracer related.  Let's start tackling this this week after we have merged @yichengt900's last PR on the budget diagnostics.

yichengt900:
@charliestock, thanks for your feedback. I agree that the second call should ultimately be placed right after the tridiagonal step, which will be addressed in PR #117. However, the focus of this PR is to remove redundant bottom calculations (e.g., o2_btm), as these diagnostics serve only as thresholds for certain sediment processes and remain unchanged before and after the sediment step within the current time step. I have also compared the diagnostic outputs between the main branch and this PR to ensure consistency. Happy to discuss further if needed.

charliestock:
I believe they are also used for the bottom diagnostics though (see cobalt_send_diagnostics), so I don't think they are redundant.  The issue seems to be that the current code only updates them after the biological source/sink calculation before sending the diagnostics.  We should updating them both for the biological source and sink and mixing/advection, but either way we have to update them, right?  Isn't this a case of moving code rather than deleting?

yichengt900:
@charliestock, thanks for your feedback. I agree that bottom diagnostics are used in `cobalt_send_diagnostics`, and PR #117 will ensure that `btm_*` are updated for both biological sources/sinks and mixing/advection.

The issue identified by @nikizadehgfdl is that some bottom diagnostic calculations, such as `btm_o2`, are duplicated **within the current sink/source codes at the same time step before sending diagnostics** and before the tridiagonal step.

Take `btm_o2` as an example, in the current main branch:

1. btm_o2 is first calculated at [line 4757](https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/f11a6b7bb99a2e3a43e649d32e6b0b07b2eea048/generic_tracers/generic_COBALT.F90#L4757) using cobalt%f_o2.
https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/f11a6b7bb99a2e3a43e649d32e6b0b07b2eea048/generic_tracers/generic_COBALT.F90#L4757

2. `cobalt%f_o2` itself is defined at [line 3085](https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/f11a6b7bb99a2e3a43e649d32e6b0b07b2eea048/generic_tracers/generic_COBALT.F90#L3085) and remains unchanged within `generic_COBALT_update_from_source` until the next time step.
https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/f11a6b7bb99a2e3a43e649d32e6b0b07b2eea048/generic_tracers/generic_COBALT.F90#L3085

3. Between [lines 4757 and 5034](https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/f11a6b7bb99a2e3a43e649d32e6b0b07b2eea048/generic_tracers/generic_COBALT.F90#L4864), `btm_o2` is used as a threshold for sediment processes, but its value remains no change.
https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/f11a6b7bb99a2e3a43e649d32e6b0b07b2eea048/generic_tracers/generic_COBALT.F90#L4864

4. The same calculation of `btm_o2` is then repeated at [line 6430](https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/f11a6b7bb99a2e3a43e649d32e6b0b07b2eea048/generic_tracers/generic_COBALT.F90#L6430) using the same `cobalt%f_o2`, before being sent to the `diagnostic manager` at [line 6572](https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/f11a6b7bb99a2e3a43e649d32e6b0b07b2eea048/generic_tracers/generic_COBALT.F90#L6572C11-L6572C32).
https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/f11a6b7bb99a2e3a43e649d32e6b0b07b2eea048/generic_tracers/generic_COBALT.F90#L6430

Since `cobalt%f_o2` does not change during the same time step in `generic_COBALT_update_from_source`, both calculations of `btm_o2` produce identical values. I confirmed this by comparing diagnostic outputs and model results before and after this PR, as well as running a global test case, which showed identical results.

It is important to note that this issue involves only four bottom diagnostics: `btm_o2`, `btm_co3_sol_calc`, `btm_co3_ion`, and `btm_omega_calc`. This PR is primarily a code cleanup to remove redundant calculations in `generic_COBALT_update_from_source` subroutine, while PR #117 will address the update timing issue you mentioned.

Perhaps I have overlooked or missed something, but any comments/suggestions are more than welcome.

charliestock:
Ahhh...I think the reason for my confusion is that this could should ultimately p_o2 and similarly update co3 and the saturation states.  I am OK with cleaning it up and then we can tackle the refinements in when we save the tracers all at once.

charliestock:
Hi @yichengt900.  Just checking in on the status of this PR.  Was it addressed when we re-organized the diagnostics?

yichengt900:
> Hi @yichengt900. Just checking in on the status of this PR. Was it addressed when we re-organized the diagnostics?

Thanks for checking, @charliestock . Yes I believe this issue has been resolved in the re-organized PR. Will close this PR for now.

PR #130, opened 2025-03-17
Add comments and show how to use Press et.al. algorithm in vertdiff_G

nikizadehgfdl:
This update
- Adds comments to vertdiff_G subroutine based on Charlie's document on it
- Adds Press et. al. tridiagonal solver subroutine for future use (possible in CBED module)
- Shows how to switch to using Press et.al. instead of the current vertical diffusion algorithm just for comparing the two algorithms.  
- Does not change current answers


generic_tracers/generic_tracer_utils.F90
```diff
@@ -3215,6 +3215,92 @@ subroutine g_tracer_vertdiff_G(g_tracer, h_old, ea, eb, dt, kg_m2_to_H, m_to_H,
...
+    !
+    ! In the present application
+    !    r_k is the old (before update) tracer concentration field at level k
+    !    u_k is the new (after  update) tracer concentration field at level k
+    !    a_k = -ea(k)/h_old(k)
+    !    c_k = -eb(k)/h_old(k)
+    !    b_k = (h_old(k)+ea(k)+eb(k))/h_old(k)
+    !
+    ! The modifications of the original algorithm is to allow for surface and buttom fluxes
+    ! as well as possible sinking of the tracers.
```
charliestock: maybe note that this can handle sinking or swimming?

generic_tracers/generic_tracer_utils.F90
```diff
@@ -3215,6 +3215,92 @@ subroutine g_tracer_vertdiff_G(g_tracer, h_old, ea, eb, dt, kg_m2_to_H, m_to_H,
...
+    !
+    ! In MOM6 models, all tracers are advected (both horizontally and vertically) by MOM6
+    ! and also horizontally diffused by MOM6 if they are registered as MOM6 tracers
+    ! (which is the case for generic tracers).
+    ! So there remains a need to vertically diffuse tracers separately by tracer packages
```
charliestock: vertically diffuse (and, if necessary vertically advect tracers due to sinking or swimming)?

generic_tracers/generic_tracer_utils.F90
```diff
@@ -3215,6 +3215,92 @@ subroutine g_tracer_vertdiff_G(g_tracer, h_old, ea, eb, dt, kg_m2_to_H, m_to_H,
...
+    !
+    ! This algorithm can be extended for the cases that the tracers have a surface or buttom
+    ! flux. In such cases the length-equivalent of tracer mass entered  to the top or buttom
+    ! layer can simply be added (with the sign of flux considered) to the right hand side
+    ! of equation (3) for k=1 and k=N equations (i.e., hI*tI_1 and hI*tI_N) .
+    !
+    !
+    ! So vertdiff is to fill the void of subgrid scale processes to correct the
+    ! vertical advection equations for diffusive processes.
+    ! We could try to include the sinking of tracers (due to gravity or active processes?)
```
charliestock: I think we already do this, right?  Maybe we could modify this comment to reflect that sinking and swimming are included and discuss this generalization?  I would be happy to help.  Maybe we could mention the modified values of the coefficients derived here:

https://docs.google.com/document/d/1ZVvLflMvTC0o9LVxQvbVSjDoQrCRqhypUNPBblwp14U/edit?tab=t.0

theresa-cordero:
@jessluo @charliestock 
Can we approve and merge this PR?

generic_tracers/generic_tracer_utils.F90
```diff
@@ -3361,6 +3459,24 @@ subroutine g_tracer_vertdiff_G(g_tracer, h_old, ea, eb, dt, kg_m2_to_H, m_to_H,
    endif
   end subroutine g_tracer_vertdiff_G
 
+  subroutine tridag_solver_Press_et_al(a,b,c,r,u,n)
```
jessluo: can we add an i to tridiag, changing this to `tridiag_solver_Press_et_al` ?

jessluo: this would also need to be changed on line 3430 and 3478.

jessluo (approved):
This looks good, sorry for not doing this earlier. I added a comment about correcting the spelling of tridiag but otherwise it looks good to me. I'm going to approve it now, that should still allow you to make the spelling change, right?

generic_tracers/generic_tracer_utils.F90
```diff
@@ -3215,6 +3215,92 @@ subroutine g_tracer_vertdiff_G(g_tracer, h_old, ea, eb, dt, kg_m2_to_H, m_to_H,
     ! concentrations after the dual-entrainments, and possibly sinking or surface
     ! and bottom sources, are applied.  The sinking is implemented with an
     ! fully implicit upwind advection scheme.
+    !
+    ! This subroutine implements a modified version of the classic tridag algorithm
```
charliestock: tridiag rather than tridag throughout?

generic_tracers/generic_tracer_utils.F90
```diff
@@ -3215,6 +3215,92 @@ subroutine g_tracer_vertdiff_G(g_tracer, h_old, ea, eb, dt, kg_m2_to_H, m_to_H,
...
+    ! from the Numerical Recepies book by Press et,al (provided below for comparison).
```
charliestock: Recipes rather than Recepies

generic_tracers/generic_tracer_utils.F90
```diff
@@ -3215,6 +3215,92 @@ subroutine g_tracer_vertdiff_G(g_tracer, h_old, ea, eb, dt, kg_m2_to_H, m_to_H,
...
+    ! -ea*tF_{k-1} + (hI+ea+eb)*tF_k -eb*tF_{k+1} = hI*tI_k  (3)
+    !
+    ! This is a tridiagonal system for unknown vector tF:
+    !
+    !  a_k*tF_{k-1} + b_k*tF_k c_k*tF_{k+1} = tI_k    k=1,...,N
+    !    a_k = -ea(k)/h_old(k)
+    !    c_k = -eb(k)/h_old(k)
+    !    b_k = (h_old(k)+ea(k)+eb(k))/h_old(k)
+    !
+    ! This algorithm can be extended for the cases that the tracers have a surface or buttom
```
charliestock: bottom rather than buttom?  (see line 3285 as well)

generic_tracers/generic_tracer_utils.F90
```diff
@@ -3215,6 +3215,92 @@ subroutine g_tracer_vertdiff_G(g_tracer, h_old, ea, eb, dt, kg_m2_to_H, m_to_H,
...
+    ! flux. In such cases the length-equivalent of tracer mass entered  to the top or buttom
+    ! layer can simply be added (with the sign of flux considered) to the right hand side
+    ! of equation (3) for k=1 and k=N equations (i.e., hI*tI_1 and hI*tI_N) .
+    !
+    !
+    ! So vertdiff is to fill the void of subgrid scale processes to correct the
+    ! vertical advection equations for diffusive processes.
```
charliestock: Perhaps note that is also handles advection relative to the fluid flow (i.e., sinking or swimming?)

charliestock (changes requested):
Hi Niki,

Very sorry letting this sit.  I miss having YC ping me each week to move these along!  I think this looks good.  I just caught a couple of typos and suggested a few places to mention sinking or swimming (i.e., vertdiff also allows us to capture any vertical motion relative to the fluid flow.

Thanks!
Charlie

generic_tracers/generic_tracer_utils.F90
```diff
@@ -48,7 +48,7 @@ module g_tracer_utils
 !-----------------------------------------------------------------------
 
   character(len=48), parameter :: mod_name = 'g_tracer_utils'
-
+  logical :: use_PressEtAl=.false.
```
andrew-c-ross: Do we also need a namelist check to turn this on/off, or is it not intended for general use yet?

generic_tracers/generic_tracer_utils.F90
```diff
@@ -3298,17 +3383,20 @@ subroutine g_tracer_vertdiff_G(g_tracer, h_old, ea, eb, dt, kg_m2_to_H, m_to_H,
...
           sink(1) = 0.0 ; h_minus_dsink(1) = (h_old(i,j,1) + sink(2))
 
-          ! Now solve the tridiagonal equation for the tracer concentrations.
+          !Prepare surface and bottom fluxes
+          sfc_src = 0.0 ; btm_src = 0.0
+          if (_ALLOCATED(g_tracer%stf)) sfc_src = (g_tracer%stf(i,j)*dt)*kg_m2_to_H
+          if (_ALLOCATED(g_tracer%btf)) btm_src = (-g_tracer%btf(i,j)*dt)*kg_m2_to_H
 
+          ! Now solve the tridiagonal equation for the tracer concentrations.
+          if(.NOT. use_PressEtAl) then !GOLD algorithm
```
andrew-c-ross: Indent the lines within this block

generic_tracers/generic_tracer_utils.F90
nikizadehgfdl: @andrew-c-ross  Since there are no namelists read in generic_tarcer_utils.F90, I could pass this as an optional argument to g_tracer_vertdiff_G and put a namelist in the caller generic_tracer.F90 to pass it through. What do you think?

generic_tracers/generic_tracer_utils.F90
nikizadehgfdl: Done. Thanks.

generic_tracers/generic_tracer_utils.F90
jessluo: I think this sounds reasonable.

andrew-c-ross: That makes sense to me as well

generic_tracers/generic_tracer_utils.F90
nikizadehgfdl: @charliestock , I pushed in updates. Please take a look. I tried to address this comment, but would be great if you can help with this item further.

andrew-c-ross (approved):

charliestock (approved):
Thanks Niki,

The changes look good to me.  Let's move ahead with this.  We can always make further additions/refinements to the comments and documentation later on if we want to.

Thanks!

nikizadehgfdl:
@jessluo, @charliestock    I have a multi-year run of a OM5-COBALT using the new Press algorithm in vertdiffG. Of course the answers change compared to not using it. What is the best diagnostics  to look at to ensure the new algorithm is doing OK and does no harm? Some vertical profiles?

Also, I see almost no difference  in the timings whether using Press subroutine or not (actually , the model runs slightly faster when I do use it).

charliestock:
Thanks @nikizadehgfdl.  I think some vertical chlorophyll profiles would probably be useful to look at.  A good place to compare would be the profile at the Hawaii Ocean Time Series, Lat/Lon = 22.75N/158W.  If you share the paths I can also look at a few additional checks.

If I recall, we had also tested this in the 1D case and in matlab?

nikizadehgfdl:
For reference the history files for the 3rd year runs without and with the Press algorithm are at:
```
/ptmp/Niki.Zadeh/archive/Niki.Zadeh/fre_om5/FMS2026.01.01_om5_20250922/om5_b12_nonbouss_cobaltv3/gfdl.ncrc5-inteloneapi252-prod/history/19600101.nc/19600101.ocean_cobalt_omip_tracers_month_z.nc
/ptmp/Niki.Zadeh/archive/Niki.Zadeh/fre_om5/FMS2026.01.01_om5_20250922/om5_b12_nonbouss_cobaltv3_pressEtAl/gfdl.ncrc5-inteloneapi252-prod/history/19600101.nc/19600101.ocean_cobalt_omip_tracers_month_z.nc
```

Here's what I get for CHL profiles at 22.75N/158W

880ac115-e07f-4b77-912d-ac8f476197cf

nikizadehgfdl:
And here's the profile of a tracer with sinking

 
a6165493-d72f-4432-8567-8418f33d7256

charliestock:
This looks solid to me Niki.  Minor differences are likely unavoidable given that the two algorithms have a different sequence of calculations etc.  Let's spend a bit of time discussing during the next COBALT doc&dev meeting to see if there are any other suggestions for further testing.  @jessluo

PR #133, opened 2025-04-02
re-organize diagnostic timing, improve formatting

charliestock:
This PR is the first step in clean-up of the diagnostic codes for COBALTv3 to support CMIP6 and CEFI.  The major task was to re-organize the “g_send_data” calls so that the diagnostics associated with the prognostic tracers, diagnostic tracers, and sinking were saved after vertdiff.  

This is essentially an expansion of the structure established in @yichengt900's PR #117, which moved the water column tracer integrals after the tridiagonal vertdiff solver with the flag “is_postvertdiff”.  The fluxes used to calculate the biological sources/sink tendencies and the associated limitations are still saved at the end of ocean_cobalt_update_from_source.

The movement of these diagnostics ensures time-step scale consistency of all tracer values and fluxes within MOM6-COBALT (i.e., the fluxes reflect those applied during the time step, prognostic tracers and diagnostics derived from them reflect values at the end of the time step). 

For carbon system variables derived from mocsy, alignment with prognostic tracer values at the end of the timestep strictly requires recalculation of the carbon chemistry after the vertdiff call.  This pull request introduces a “recalculate_carbon” logical to the cobalt type with a default value of .true. .  If .false., the recalculation is skipped and the carbon system values kept at the beginning of update_from_source are kept.

The addition of carbon and 100m integrals to cobalt_send_diagnostics required an expansion of the arguments and routines available to the cobalt_send_diagnostics routine.  This in turn required small changes to generic_tracer.F90 and MOM_generic_tracer.F90.

In addition to refining the timing of the diagnostic sends, this PR also cleans up a number of other elements of the diagnostic code:

- The formatting of the calls has been cleaned up and made uniform.
- 3D fluxes that were saved as layer integrals (units of moles m-2 sec-1 by multiplying by rho_dzt) were changed to moles kg sec-1.  The layer integral approach was a legacy of MOM4/5.  It allowed you to easily sum layers to calculate integrated fluxes in a Z* framework.  The interpretation of layer integrals is more challenging in hybrid coordinates and they really don’t make sense when re-mapped.  This simplified the flux diagnostics (and allowed us to get rid of a few).
- replaced cases where cmip diagnostics were multiplied by rho_dzt/dzt with multiplication by cobalt%Rho_0 to simplify.
- Defined nmd_misc and nlg_misc for CMIP calculations.
- Updated the following diagnostics for COBALTv3: graz, intppnitrate
- Fixed a bug in the units of the CMIP surface chlorophyll diagnostics (1e6)
- Clarified units of mesozoo_200 (general strengthening of metadata will be handled in a subsequent PR)

Additional questions for the group:

- Many lines of code of the diagnostic code are spent extracting the top grid cell for surface properties.  Could these be eliminated and the diagnostics made more consistent by defining a surface layer in the diagnostic table?
- Where in the thermodynamic time step are the actual prognostic tracer values saved?
- Should we make the 100m integrals have a user-specified integration depth (i.e., _100m = _intz where z is specified in metadata)?
- What is the best way to test/verify such a broad PR?

### DIFF
--- a/generic_tracers/cobalt_reg_diag.F90
+++ b/generic_tracers/cobalt_reg_diag.F90
@@ -353,23 +353,23 @@ module COBALT_reg_diag
     ! Register diagnostics for phytoplankton loss terms: zooplankton
     ! CAS: loss diagnostics simplified to just N
 
-    vardesc_temp = vardesc("jzloss_n_Di","Diazotroph nitrogen loss to zooplankton layer integral",&
-                           'h','L','s','mol N m-2 s-1','f')
+    vardesc_temp = vardesc("jzloss_n_Di","Diazotroph nitrogen loss to zooplankton",&
+                           'h','L','s','mol N kg-1 s-1','f')
     phyto(DIAZO)%id_jzloss_n = register_diag_field(package_name, vardesc_temp%name, axes(1:3),&
          init_time, vardesc_temp%longname,vardesc_temp%units, missing_value = missing_value1)
 
-    vardesc_temp = vardesc("jzloss_n_Lg","Large phyto nitrogen loss to zooplankton layer integral",&
-                           'h','L','s','mol N m-2 s-1','f')
+    vardesc_temp = vardesc("jzloss_n_Lg","Large phyto nitrogen loss to zooplankton",&
+                           'h','L','s','mol N kg-1 s-1','f')
     phyto(LARGE)%id_jzloss_n = register_diag_field(package_name, vardesc_temp%name, axes(1:3),&
          init_time, vardesc_temp%longname,vardesc_temp%units, missing_value = missing_value1)
 
-    vardesc_temp = vardesc("jzloss_n_Md","Medium phyto nitrogen loss to zooplankton layer integral",&
-                           'h','L','s','mol N m-2 s-1','f')
+    vardesc_temp = vardesc("jzloss_n_Md","Medium phyto nitrogen loss to zooplankton",&
+                           'h','L','s','mol N kg-1 s-1','f')
     phyto(MEDIUM)%id_jzloss_n = register_diag_field(package_name, vardesc_temp%name, axes(1:3),&
          init_time, vardesc_temp%longname,vardesc_temp%units, missing_value = missing_value1)
 
-    vardesc_temp = vardesc("jzloss_n_Sm","Small phyto nitrogen loss to zooplankton layer integral",&
-                           'h','L','s','mol N m-2 s-1','f')
+    vardesc_temp = vardesc("jzloss_n_Sm","Small phyto nitrogen loss to zooplankton",&
+                           'h','L','s','mol N kg-1 s-1','f')
     phyto(SMALL)%id_jzloss_n = register_diag_field(package_name, vardesc_temp%name, axes(1:3),&
          init_time, vardesc_temp%longname,vardesc_temp%units, missing_value = missing_value1)
 
@@ -377,23 +377,23 @@ module COBALT_reg_diag
     !  Register diagnostics for phytoplankton loss terms: aggregation
     !
 
-    vardesc_temp = vardesc("jaggloss_n_Di","Diazotroph nitrogen loss to aggregation layer integral",&
-                           'h','L','s','mol N m-2 s-1','f')
+    vardesc_temp = vardesc("jaggloss_n_Di","Diazotroph nitrogen loss to aggregation",&
+                           'h','L','s','mol N kg-1 s-1','f')
     phyto(DIAZO)%id_jaggloss_n = register_diag_field(package_name, vardesc_temp%name, axes(1:3),&
          init_time, vardesc_temp%longname,vardesc_temp%units, missing_value = missing_value1)
 
-    vardesc_temp = vardesc("jaggloss_n_Lg","Large phyto nitrogen loss to aggregation layer integral",&
-                           'h','L','s','mol N m-2 s-1','f')
+    vardesc_temp = vardesc("jaggloss_n_Lg","Large phyto nitrogen loss to aggregation",&
+                           'h','L','s','mol N kg-1 s-1','f')
     phyto(LARGE)%id_jaggloss_n = register_diag_field(package_name, vardesc_temp%name, axes(1:3),&
          init_time, vardesc_temp%longname,vardesc_temp%units, missing_value = missing_value1)
 
-    vardesc_temp = vardesc("jaggloss_n_Md","Medium phyto nitrogen loss to aggregation layer integral",&
-                           'h','L','s','mol N m-2 s-1','f')
+    vardesc_temp = vardesc("jaggloss_n_Md","Medium phyto nitrogen loss to aggregation",&
+                           'h','L','s','mol N kg-1 s-1','f')
     phyto(MEDIUM)%id_jaggloss_n = register_diag_field(package_name, vardesc_temp%name, axes(1:3),&
          init_time, vardesc_temp%longname,vardesc_temp%units, missing_value = missing_value1)
 
-    vardesc_temp = vardesc("jaggloss_n_Sm","Small phyto nitrogen loss to aggregation layer integral",&
-                           'h','L','s','mol N m-2 s-1','f')
+    vardesc_temp = vardesc("jaggloss_n_Sm","Small phyto nitrogen loss to aggregation",&
+                           'h','L','s','mol N kg-1 s-1','f')
     phyto(SMALL)%id_jaggloss_n = register_diag_field(package_name, vardesc_temp%name, axes(1:3),&
          init_time, vardesc_temp%longname,vardesc_temp%units, missing_value = missing_value1)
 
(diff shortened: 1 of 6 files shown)


yichengt900:
@charliestock, thanks for opening this PR — it's an important first step in consolidating the diagnostics.

This PR appears to significantly reorganize the diagnostics code, particularly around `g_send_data` calls. While much of the change involves moving code around, it’s still a large diff and will take me some time to carefully review. On first look, this looks like a clean and well-documented PR.

**I’ve updated the if statement logic for `recalculate_carbon` so that the PR now passes the 1D CI test.**

**I've found another bug in `get_param` call for `recalculate_carbon`.**

**I’m okay with the `recalculate_carbon` option and keeping the default as true to ensure our diagnostic output remains as accurate as possible. However, we’ll need to check whether it has any significant impact on computational time. My impression is that it probably doesn’t, but I’ll run some test cases to confirm.**

I also took this opportunity to audit all `id_*` diagnostics to check whether they are properly registered and used via `g_send_data`. I found several issues: some IDs are registered but never sent, some are `used in g_send_data` without being registered, and some are neither registered nor used anywhere in the code. These IDs (and associated diagnostic outputs) are currently non-functional, and this PR seems like a good opportunity to either clean them up or fix their usage. I have summarized them below:

**IDs that are registered but never called by `g_send_data`:**
`id_chlmiscos`
`id_det_jhploss_n`
`id_det_jzloss_n`
`id_f_n_100`
`id_jdiss_cadet_arag_plus_btm`
`id_jdiss_cadet_calc_plus_btm`
`id_sfc_sio4`
`id_thetao`
`id_zo2min`
`id_zsatarag`
`id_zsatcalc`

**IDs that are called by `g_send_data` without being registered:**
 `id_co2_alpha`
`id_co2_csurf`
`id_f_alk_int_100`
`id_f_dic_int_100`
`id_f_din_int_100`
`id_f_fed_int_100`
`id_f_po4_int_100`
`id_f_sio4_int_100`
`id_grid_kmt_diag`
`id_irr_aclm_z`
`id_jalk_100`
`id_jdic_100`
`id_jdin_100`
`id_jfed_100`
`id_jpo4_100`
`id_jsio4_100`
`id_k_bot_diag`
`id_kmt_diag`
`id_nh3_alpha`
`id_nh3_csurf`
`id_no3_in_source`
`id_pco2_csurf`
`id_rho_dzt_bot_diag`
`id_rho_dzt_kmt_diag`

**IDs that are never used:**:
`id_def_p`
`id_jaggloss_fe`
`id_jaggloss_p`
`id_jaggloss_sio2`
`id_jexuloss_fe`
`id_jexuloss_p`
`id_jhploss_fe`
`id_jhploss_p`
`id_jhploss_sio2`
`id_jmortloss_fe`
`id_jmortloss_p`
`id_jmortloss_sio2`
`id_jo2c`
`id_jvirloss_fe`
`id_jvirloss_p`
`id_jvirloss_sio2`
`id_jzloss_fe`
`id_jzloss_p`
`id_jzloss_sio2`
`id_nlg_misc`
`id_nmd_misc`

Some thoughts on the questions:

1. Yes, I think that’s possible. We could use the `sub_region` output feature in the diag_manager and restrict output to just the surface layer.

2. If you’re asking when the prognostic tracer diagnostics call the diag_manager and get saved (as defined in `diag_table`), it happens after the `generic_tracer_update_from_bottom` call in [MOM_generic_tracer.F90](https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/9e64a3bca6a22f4fcba60988c48fe196ed4b8d3c/generic_tracers/MOM_generic_tracer.F90#L866).

3. This seems doable. We could follow the similar approach like the user-specified [MLD diagnostics](https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/MOM6/blob/9bd3bfd11990752f5d57b37ef68b1881d25c1fd8/src/parameterizations/vertical/MOM_diabatic_driver.F90#L492) and allow for a user-specified integration depth.

4. Good question. At this stage, we don’t have a solid baseline for verifying diagnostics output. However, this PR provides a good opportunity to establish one — perhaps by archiving the diagnostics output from the 1D test case. That could serve as a reference for future PRs.

generic_tracers/cobalt_send_diag.F90
```diff
@@ -48,2406 +71,1818 @@ subroutine cobalt_send_diagnostics(tracer_list, model_time,grid_tmask,Temp,rho_d
...
+              cobalt%co3_sol_arag(i,j,k) = cobalt%f_co3_ion(i,j,k) / max(cobalt%omega_arag(i,j,k),epsln)
+              cobalt%co3_sol_calc(i,j,k) = cobalt%f_co3_ion(i,j,k) / max(cobalt%omega_calc(i,j,k),epsln)
+              ! Update diatom and misc phytoplankton groups for diagnostics
+              cobalt%nlg_diatoms(i,j,k)=phyto(LARGE)%f_n(i,j,k)*phyto(LARGE)%silim(i,j,k)
+              cobalt%nmd_diatoms(i,j,k)=phyto(MEDIUM)%f_n(i,j,k)*phyto(MEDIUM)%silim(i,j,k)
+              cobalt%nlg_misc(i,j,k)=phyto(LARGE)%f_n(i,j,k) - phyto(LARGE)%f_n(i,j,k)*phyto(LARGE)%silim(i,j,k)
+              cobalt%nmd_misc(i,j,k)=phyto(MEDIUM)%f_n(i,j,k) - phyto(MEDIUM)%f_n(i,j,k)*phyto(MEDIUM)%silim(i,j,k)
+            enddo; enddo ; enddo !} i,j,k
+
+          endif !} recalculate carbon system properties
```
yichengt900, line 192: Preliminary results show that the carbon system recalculation has minimal impact on runtime (< 0.5%), so we should be good. One quick thought—although this is just a small portion of the code, we might consider turning it into a subroutine or function if we plan to reuse this calculation in multiple places.

generic_tracers/cobalt_reg_diag.F90
```diff
@@ -502,460 +502,459 @@ subroutine cobalt_reg_diagnostics(diag_list,axes,init_time,phyto,zoo,bact,cobalt
     cobalt%id_nmd_diatoms = register_diag_field(package_name, vardesc_temp%name, axes(1:3),&
          init_time, vardesc_temp%longname,vardesc_temp%units, missing_value = missing_value1)
 
-    vardesc_temp = vardesc("q_si_2_n_lg_diatoms","Si:N ratio in large diatoms",&
-                           'h','L','s','mol Si mol N','f')
-    cobalt%id_q_si_2_n_lg_diatoms = register_diag_field(package_name, vardesc_temp%name, axes(1:3),&
+    vardesc_temp = vardesc("nlg_misc","large phytoplankton nitrogen from misc non-diatoms",&
+                           'h','L','s','mol kg-1','f') 
+    cobalt%id_nlg_diatoms = register_diag_field(package_name, vardesc_temp%name, axes(1:3),&
```
yichengt900, line 867: `id_nlg_misc `?

generic_tracers/cobalt_reg_diag.F90
```diff
@@ -502,460 +502,459 @@ subroutine cobalt_reg_diagnostics(diag_list,axes,init_time,phyto,zoo,bact,cobalt
...
          init_time, vardesc_temp%longname,vardesc_temp%units, missing_value = missing_value1)
 
-    vardesc_temp = vardesc("q_si_2_n_md_diatoms","Si:N ratio in medium diatoms",&
-                           'h','L','s','mol Si mol N','f')
-    cobalt%id_q_si_2_n_md_diatoms = register_diag_field(package_name, vardesc_temp%name, axes(1:3),&
+    vardesc_temp = vardesc("nmd_misc","medium phytoplankton nitrogen from misc. non-diatoms",&
+                           'h','L','s','mol kg-1','f')
+    cobalt%id_nmd_diatoms = register_diag_field(package_name, vardesc_temp%name, axes(1:3),&
```
yichengt900, line 872: `id_nmd_misc `?

charliestock:
Thanks for the recalculate_carbon improvements and fixes @yichengt900.   

I have also gone through the variables identified in your id audit and addressed as follows in the latest commit:

**Registered but never called:**
id_chlmiscos: Accidentally ended up sending this to id_chldiatos after the copy and pastes associated with this PR.  This commit now correctly sends miscellaneous phytos to id_chlmiscos.
id_det_jhploss_n, id_det_jzloss_n: Not active, but I have added as a placeholder in case detrivory is activated.
id_f_n_100: Missed while moving things around.  Added 100m for phyto, zoo and bacteria to cobalt_send_diag.F90.
id_jdiss_cadet_arag_plus_btm, id_jdiss_cadet_calc_plus_btm: Added but not sure we will ultimately need.
id_sfc_sio4: missed when moving things around.  Corrected.
id_thetao: Perhaps better saved in the ocean model, but I added back the option to save in BGC as well and not cause any problems with existing diag tables
id_zo2min, id_zsatarag, id_zsatcalc: Removed  duplicates id_z_o2min, id_z_sat_arag, id_z_sat_calc and consolidated code around zo2min zsatarag and zsatcalc, which I believe come from CMIP.  Still need to confirm that these diagnostics are working properly.

**Called without being registered:** added registration for all variables except id_no3_in_source and id_pco2_csurf.  We don't need "no3_in_source" and "pco2_csurf" is redundant is the "pco2surf" diagnosic.  I also added a register entry for irr_aclm and I could not find id_kmt_diag.

**IDs that are never used**: added registrations and sends for all, with the exception of id_def_p and id_jo2c, which are not needed.  The addition of diagnostics for all of the loss terms on a nutrient-by-nutrient level does add quite a few new registrations and g_send_data calls.  Do these only impact the speed of the run if the diag table wants them?

yichengt900:
@charliestock, thanks for the new updates! It looks like most of the variables are both registered and likely called by `g_send_data`, except for the two newly introduced variables, `id_nlg_misc` and `id_nmd_misc` (please check my previous single comments). I also checked the timing again, and it seems that it was not affected if we do not add them to the `diag_table`. Overall, I think this PR is in good shape, but it will need another round of checks. One thing we could consider is establishing baseline diagnostics based on this PR, so we can use them as a reference for future PRs.

yichengt900:
Ok @charliestock, I ran two 0.5° 5-year runs using the original diagnostic code from the current branch and the new code from this PR. The outputs are now available on PPAN:
```
/archive/ynt/fre/OM4_05/2025_04_diag_new_comp/OM4_05_COBALTv3_jra55_const_recalc_diag_new
/archive/ynt/fre/OM4_05/2025_04_diag_org_comp/OM4_05_COBALTv3_jra55_const_recalc_diag_org
```

I briefly checked a few outputs like `intpp` and `ph(f_htotal)`, and they look good to me:

`intpp` (left: current codes; right: new codes from this branch):
Screenshot 2025-04-16 at 9 53 43 AM

`ph (f_htotal; left: current codes; right: new codes from this branch)`:
Screenshot 2025-04-16 at 10 19 58 AM

I may have missed something, but overall everything looks ok to me. I’d appreciate it if someone else could take a deeper look before we move forward with merging this PR.

charliestock:
Hello @yichengt900, thank you!  I have been taking a deeper look and I think that things look good so far.  The only thing I noticed that is a bit interesting is that saving the tracers after the tridiagonal leads to slightly lower surface values for some quanitities.  For example, phycos is about 2% higher in regions with large phytoplankton when quantities are saved before the tridiag because they haven't had a chance to sink (see below).  This is larger for detocos, but doesn't change the overall patterns in a significant way (detoc varies by 2 orders of magnitude) and the new way is more correct.

I will continue to look through these over the weekend, but I think we should move forward with the merge given the current uncertainties.  

Screenshot from 2025-04-18 12-41-10

Screenshot from 2025-04-18 13-18-01

yichengt900:
@charliestock, thanks for the detailed feedback! It makes sense that the values would be slightly higher before the tridiagonal solver, as the sinking effect hasn’t been applied yet. I’ll go ahead and merge the PR first and make a backup for reference. Thanks again!

yichengt900 (approved):

charliestock:
Sound great YC, thank you!  I will try and do some more checks over the weekend.

PR #137, opened 2025-05-12
add acclimation memory for temp and nut lims (testing)

charliestock:
Last week, @fabienpaulot noticed that COBALT can generate significant diel chlorophyll variability in strongly stratified regions.  In the strongest cases, the range of variation was comparable to or slightly greater than the mean chlorophyll value.  This is illustrated in the figure below, which shows the standard deviation of 24 hours of surface chlorophyll variability divided by the mean chlorophyll: 

Diel_Chl_Variability

We were surprised by this because the light forcing used for photoacclimation is low-pass filtered to damp diurnal cycles.  After a bit of digging, however, we found that the chlorophyll variability was driven by diel variations in nutrient limitation: Nutrients in the subtropics would build up modestly at night and rapidly deplete the next morning.  In short, there is a significant diel cycle in surface nutrient limitation in highly stratified regions.

While there is evidence for significant diel chlorophyll variability of this kind, e.g.:

https://doi.org/10.1016/j.ecss.2006.05.019
https://doi.org/10.1029/2000JC000747
https://doi.org/10.1080/01431161.2018.1538583

It is likely too strong.  This PR thus allows for the filtering of the nutrient and temperature limitation product in a manner consistent with the light limitation.  The plot below compares the diel chlorophyll variability in the original formulation (left) to that with 1 day (middle) and 3 day (right) acclimation timescales:

Aclm_comparison

The damping of variability at longer timescales is clear.  In the 3 day case, however, there is an interesting increase in the equatorial Pacific variability.  This was because carbon-driven variability is partly masked by compensating photoacclimation with shorter acclimation time-scales, but shows through at longer time scales.

While the diurnal chlorophyll variability is quite significant in nutrient poor regions, the variations above have minimal impacts on the mean chlorophyll, NPP or other properties because the areas impacted are all at light saturating levels.  The significant changes in chlorophyll are just sliding the growth response a bit further up or down an asymptotic curve with minimal impact on growth or the daily mean chlorophyll.  The response, however, does create a slight shoaling of the deep chlorophyll maximum in impacted areas since the mean chlorophyll during daylight hours increases slightly.

Here is a plot of the mean chlorophyll and DCM impact at the Hawaii Ocean Time Series.  The top row is the mean chlorophyll for the three cases (original, 1 day for nutrient*temp, 3 day for nutrient*temp).  Bottom right is the DCM response at HOT in August, and bottom middle and right are the ratios of the perturbations to the original:

Chl_and_DCM_sensitivity_to_acclim

And here is a similar plot for NPP:

NPP_sensitivity_to_acclim
 
Thanks to @fabienpaulot for raising this issue and to @fabienpaulot and @jessluo for conversations that helped resolve it.


generic_tracers/generic_COBALT.F90
```diff
@@ -3528,25 +3584,32 @@ subroutine generic_COBALT_update_from_source(tracer_list,Temp,Salt,rho_dzt,dzt,h
...
+          endif 
           do m = 1,cobalt%numlightadapt
             ! since we test the low and high, divide by m-1 so first step is the low and last is the high
             alpha_step = (phyto(n)%alpha_ll - phyto(n)%alpha_hl)/(real(cobalt%numlightadapt,8)-1.0)
             alpha_temp = phyto(n)%alpha_hl + (real(m,8)-1.0)*alpha_step
             P_C_max_step = (phyto(n)%P_C_max_hl - phyto(n)%P_C_max_ll)/(real(cobalt%numlightadapt,8)-1.0)
             P_C_max_temp = phyto(n)%P_C_max_hl - (real(m,8)-1.0)*P_C_max_step
-            P_C_m_temp = max(phyto(n)%liebig_lim(i,j,k)*P_C_max_temp*cobalt%expkT(i,j,k),epsln)
-            theta_temp = max(phyto(n)%thetamax/(1.0 + phyto(n)%thetamax*alpha_temp*cobalt%f_irr_aclm(i,j,k)*0.5/P_C_m_temp), &
+            P_C_m_aclm = max(P_C_max_temp*phyto(n)%f_pcmlim_aclm(i,j,k),epsln)
```
yichengt900, line 3643: While the new form is mathematically identical to the old one, floating-point operations are not associative, which caused slight differences in the results. One simple solution is the recalculate `P_C_m_aclm` when `photoaclm_opt=0`:
```
(cobalt%photoaclm_opt.eq.0) P_C_m_aclm =max(phyto(n)%liebig_lim(i,j,k)*P_C_max_temp*cobalt%expkT(i,j,k),epsln)
```
if we really want to preserve the answer here. Since we are **okay** with changes in the answers at this point, I would say we can let this one pass for now, as the changes look mathematically sound to me. I will have update the reference accordingly.

generic_tracers/generic_COBALT.F90
```diff
@@ -2471,6 +2473,30 @@ subroutine user_add_tracers(tracer_list)
...
+
+     call g_tracer_add(tracer_list,package_name,&
+         name       = 'pcmlim_aclm_nmd',    &
+         longname   = 'Nut*Temp Lim memory, medium phytoplankton', &
+         units      = 'dimensionless',         &
+         prog       = .false.              )
+
+     call g_tracer_add(tracer_list,package_name,&
+         name       = 'pcmlim_aclm_nlg',    &
+         longname   = 'Nut*Temp Lim memory for photosynthesis, large phytoplankton', &
```
yichengt900: Can we make the descriptions of the long names consistent with each other?

generic_tracers/generic_COBALT.F90
charliestock: Thanks @yichengt900.  I believe your suggested change would also allow us to take out the if statement on line 3590.  If we can be bit wise reproducible without any added complexity, we should probably got for it.  I will make this change.

generic_tracers/generic_COBALT.F90
charliestock: Will do, thanks @yichengt900.

charliestock (commented):
Meant to just submit comments, I will work on a commit to address the issues you have raised @yichengt900.

yichengt900 (approved):
@charliestock, thanks for the updates. This PR has now passed regression testing and the changes look good to me. Approved.

PR #140, opened 2025-05-13
fixed  btm_*  and *_100 diagnostics

yichengt900:
As titled. When #133 was introduced, it was a large PR and unfortunately, we didn’t review it thoroughly. As a result, a silly bug slipped through—mainly incorrect dimensions for a few variables such as `Temp`, `Salt`, `rho_dzt`, `dzt`and a few others.

This PR fixes that issue by passing those variables with the correct dimensions. We also simplified the interface of `cobalt_send_diagnostics` and now use `g_tracer_get_common` to retrieve all the index values and masks. Below are screenshots demonstrating that we now have the correct outputs.
Screenshot from 2025-05-13 11-11-46
Screenshot from 2025-05-13 11-13-36

### DIFF
--- a/generic_tracers/cobalt_send_diag.F90
+++ b/generic_tracers/cobalt_send_diag.F90
@@ -19,7 +19,7 @@ module COBALT_send_diag
   use time_manager_mod,  only: time_type
 
   use g_tracer_utils, only : g_send_data,g_tracer_get_pointer,g_tracer_set_values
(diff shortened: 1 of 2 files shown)


jessluo:
This looks great. I like the simplification and I'm glad the fix was relatively straightforward!

jessluo (approved):

PR #143, opened 2025-06-05
Clean-up flux calculation

fabienpaulot:
Move the calculations needed for atm/ocn flux exchange outside of COBALT in a new module "generic_coupler_fluxes".
This new module is intended to be used in AMIP to ensure as much consistency as possible between AMIP and COUPLED.
Let me know what you think 

@nikizadehgfdl @jessluo @charliestock @yichengt900


yichengt900:
Thanks @fabienpaulot. I didn’t really test your feature branch, but the code looks good to me by eye.

charliestock:
Hello @fabienpaulot and @nikizadehgfdl, thank you for putting this together.  I think it is a great direction to head in!  

A couple of questions and concerns:

1) What is the intended scope of this module?  Will it ultimately cover all the coupler fluxes - i.e., O2, CO2, dust?

2) Will this routine work for any BGC module or is it COBALT-specific?  If it is COBALT-specific, should that be reflected in the module name?

3) Could we add some high-level comments to convey the current and ultimate intended scope of the module? (e.g., we current include NH4 and DMS, but intent is to expand to other fluxes that are currently handles in....) 

4) Could we add a brief comment at the beginning of each subroutine describing the purpose of each?

5) Could we work with @theresa-cordero to fit this into the COBALT flow diagram?

Lastly, I am concerned about sneaking this in before the code freeze, but I would be open to it following a discussion and if we had rigorous testing prior to pressing go.  Looping in @jessluo and @yichengt900 for consideration.

Thanks again - I think this is definitely a move in a good direction. I'd just like to understand the big picture better and have some concerns with the timing.

best,
Charlie

fabienpaulot:
Thanks Charlie. 

Following your questions and offline discussions, I repackaged this PR to only include flux-related calculations in a new module named "ocean_atm_chemical_fluxes_utils.F90". No capability is added/removed.

This new module can be thought as a toolbox to calculate chemical properties required for atm/ocean gas exchange.
The calculations are not COBALT-specific. Currently, I only included NH3 and DMS but we could move CO2, O2 if you so wish (I was trying to be conservative!). 
The overall goal is to ensure consistency between the fully-coupled and amip configurations for ocean/atm gas exchange. 
A side benefit is to remove these calculations from COBALT. 

I have checked that the proposed changes do not impact model results (ardiff on restarts) 

Adding @jessluo and @theresa-cordero  to the discussion,

charliestock:
Thanks Fabien,

@yichengt900  and I were discussing this more today and had some additional question.

If we were to move CO2 and O2 into this subroutine, would it need to call mocsy to calculate the carbon chemistry?  Maybe we should consolidate the air-sea exchange within COBALT and then decide whether we want to move things out?  Preparing the ocean BGC tracers to be exchanged with the atmosphere is a substantial part of the ocean BGC calculation....

best, @charliestock

yichengt900:
@fabienpaulot, per our conversation, I will leave this PR open for now and keep this topic as an open discussion for future code infrastructure improvements. Thanks again!

nikizadehgfdl:
@fabienpaulot could you push the "Resolve conflicts" button and see if it goes easy? 
Alternatively (and perhaps easier), we can  close this and open a new PR to https://github.com/NOAA-GFDL/ocean_BGC/tree/esm45  to get ready for AM4.5 public release . Thanks.

fabienpaulot:
Hi Niki,
it looks like there is a single conflict (unit change). Fabein

andrew-c-ross:
I made an executive decision to solve the units merge conflict using the current dev/cefi units which were last updated by #151. #151 is described as a correction for these units.

generic_tracers/generic_COBALT.F90
andrew-c-ross: `rgdas` and `wtmair` are unused now. 

```suggestion
  use constants_mod,     only: WTMCO2, WTMO2,WTMN
```

andrew-c-ross: `phos_nh3_override` is unused now.

```suggestion  
```

generic_tracers/ocean_atm_chemical_fluxes_utils.F90
```diff
@@ -0,0 +1,197 @@
...
+              + (a4+a5*tk+a6/tk)*salt              &
+              + (a7+a8/tk)*salt**2.                &
+              + (a9+a10/tk)*salt**3.
+
+  end function calc_pka_nh3
+
+  !salting out correction for solubility (Johnson 2010, Ocean Science)
+  function saltout_correction(kh,vb,salt) result(C)
+      real, intent(in) :: Kh,vb,salt
+      real*8 :: T,log_kh,theta2
```
andrew-c-ross: `theta2` and `T` are unused.

```suggestion
      real*8 :: log_kh
```

generic_tracers/ocean_atm_chemical_fluxes_utils.F90
```diff
@@ -0,0 +1,197 @@
...
+
+  subroutine get_dms_ocean_atm_flux_property(tc,salt,f_dms,pH,rho_0, &
+       dms_alpha,dms_csurf,pdms_csurf)
+
+    real, intent(in)  :: tc,salt,f_dms,pH,rho_0
+    real, intent(out) :: dms_alpha, dms_csurf
+    real, intent(out), optional :: pdms_csurf
+
+    dms_alpha = 0.537023e3*exp(3500.*(1./(tc+273.15)-1./298.15)) !M/atm
+    dms_alpha = dms_alpha/saltout_correction(101325./(1.e-3*rdgas*wtmair*tc*dms_alpha),vb_dms,salt)*1./rho_0 !mol/m3/atm
```
andrew-c-ross: There's probably a units error here: line 26 uses `tc` directly, but line 52 uses `tc+273.15` in an equivalent expression for nh3

andrew-c-ross (changes requested):
Just needs a check for a possible units error on temperature, and cleanup of a few newly unused variables.

generic_tracers/ocean_atm_chemical_fluxes_utils.F90
fabienpaulot: Thanks Andrew. It's a mistake indeed, temperature should be in Kelvin. 
Note that the DMS exchange code has not been used (NH3 has been used).

nikizadehgfdl:
I teseted ESM4.5 with this branch and it  does not change ESM4.5 answers.

PR #144, opened 2025-06-12
remove nh3 as a tracer

fabienpaulot:
NH3 is defined a diagnostic tracer in COBALT but it's not necessary since it's recalculated at every time step from NH4. 

This PR removes NH3 from the list of diagnostic tracers. The following changes are needed in the xml:

remove: 
nh3_src_file = INPUT/init_ocean_cobalt_nh3.res.nc                                                                                                                                                                  
nh3_src_var_name = nh3                                                                                                                                                                                             
nh3_src_var_unit = none                                                                                                                                                                                            
nh3_dest_var_name = nh3                                                                                                                                                                                            
nh3_dest_var_unit = mol kg-1                                                                                                                                                                                       
nh3_src_var_record = 1                                                                                                                                                                                             
nh3_src_var_gridspec = NONE                                                                                                                                                                                        
nh3_valid_min = 0.0   

remove (optional)
      <dataFile label="input" target="INPUT/" chksum="" size="" timestamp="">                                                                                                                                      
        <dataSource site="ncrc">$(CMIP6_WORLD)/ESM4/OBGC/IC/COBALT/init_ocean_cobalt_nh3.res.nc</dataSource>                                                                                                       
      </dataFile> 

These changes do not cause any change to the model output

### DIFF
--- a/generic_tracers/generic_COBALT.F90
+++ b/generic_tracers/generic_COBALT.F90
@@ -2536,15 +2536,6 @@ contains
          units      = 'dimensionless',         &
          prog       = .false.              )
 
-    if (do_nh3_atm_ocean_exchange .or. scheme_nitrif.eq.2 .or. scheme_nitrif.eq.3) then
-       call g_tracer_add(tracer_list,package_name,&
-            name       = 'nh3',         &
-            longname   = 'NH3', &
-            units      = 'mol/kg',     &
-            prog       = .false.,       &
-            init_value = 1.e-10           )
-    end if
-
   end subroutine user_add_tracers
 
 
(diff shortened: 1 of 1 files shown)


yichengt900 (commented):
@fabienpaulot, thanks for opening this PR. As we discussed, this change only removes nh3 from the restart file, which shouldn't be a problem since it's recalculated at every time step and we don’t actually use the restart value for nh3. I'm okay with these code changes. CC @charliestock to keep him in the loop.

jessluo (approved):
Yes, this looks good to me too. Thanks Fabien! Since NH3 is not needed as a diagnostic tracer then it would be good to remove it from the code.

yichengt900 (approved):
Tested this PR further with the NEP and NWA configurations, and again, it did not change the answers. Restart files are different due to the fact that this tracer is no longer output as a diagnostic tracer.

PR #149, opened 2025-06-24
Correction for enhanced CaCO3 dissolution 

feida6996:
This update modifies the section of code that executes when do_resp_ca_diss is enabled, which applies enhanced CaCO₃ dissolution associated with organic matter remineralization. The change introduces a constraint using the local concentration of aragonite (f_cadet_arag) and calcite (f_cadet_calc), ensuring that the enhanced dissolution is physically limited by the availability of particulate CaCO₃.

Without this constraint, the remineralization-driven dissolution can become unrealistically large in regions with strong organic matter remin, even when particulate CaCO₃ is low — in some cases leading to negative concentrations of aragonite or calcite.

A figure illustrating the impact of this correction is included. The results look similar to previous simulations but now avoid unphysical negative values.

correct_enhanced_diss

### DIFF
--- a/generic_tracers/generic_COBALT.F90
+++ b/generic_tracers/generic_COBALT.F90
@@ -4933,9 +4933,11 @@ contains
     if (cobalt%do_resp_ca_diss) then
         do k=1,nk ; do j=jsc,jec ; do i=isc,iec  !{
            cobalt%jdiss_cadet_arag(i,j,k) = cobalt%jdiss_cadet_arag(i,j,k) + &
-                                            cobalt%resp_ca_2_n_arag * cobalt%jremin_ndet(i,j,k)
+                                            cobalt%resp_ca_2_n_arag * cobalt%f_cadet_arag(i,j,k) * &
+                                            cobalt%jremin_ndet(i,j,k)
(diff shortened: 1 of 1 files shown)


yichengt900 (approved):

generic_tracers/generic_COBALT.F90
yangyzhao5, line 1: @feida6996 I wonder if the introduction of cadet_arag and cadet_calc concentrations would lead to the inconsistency of the units between left and right sides.

charliestock: Thanks @yangyzhao5.  The PR does change the units of the "resp_ca_2_n" constants, so we should indicate this in generic_COBALT.F90.  If I am not mistaken the units should now be "(mole N kg)-1" since you essentially divide the prior constant - which was in "moles calc (moles N)-1" by "moles Calc kg-1".  Do we agree?  If so, @feida6996, could you do a quick PR to make this change?

feida6996: Thank you @yangyzhao5 for the reminder. Yes @charliestock , as we discussed before the new unit should be kg (mol N)-1. I will open the PR now. Sorry for missing this!

PR #148, opened 2025-06-20
Nitrate constraint on anaerobic remineralization of DOM by bacteria to avoid negative nitrate values

yangyzhao5:
This PR is to add nitrate constraint on anaerobic remineralization of dissolved organic material by bacteria, similar to on that of particulate organic material, to avoid nitrate consumption in low-oxygen conditions when nitrate concentrations is reduced to zero (especially when anammox is switched on and fixed nitrogen loss is enlarged).

PR #150, opened 2025-06-25
Enhance logic for no3lim

yichengt900:
PR #148 adds a nitrate constraint to prevent nitrate consumption under low-oxygen conditions. This new logic introduces a slight change to our NWA and NEP baselines.

@charliestock, as I recall, we didn’t see negative nitrate concentrations when anammox was off (`cobalt%gamma_nh4amx` == 0), correct? This PR simply enhances the logic for the `no3lim` term, making it active only when anammox is on and under low-oxygen conditions. By doing this, we can preserve the baseline for our CEFI regional applications.

### DIFF
--- a/generic_tracers/generic_COBALT.F90
+++ b/generic_tracers/generic_COBALT.F90
@@ -4019,7 +4019,7 @@ contains
                               (cobalt%k_o2 + max(cobalt%f_o2(i,j,k),cobalt%o2_min))
        ! Note that nitrate availability affects the anaerobic remineralization of dissolved organic material 
        ! as that of particulate organic material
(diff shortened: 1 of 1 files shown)


charliestock:
Hi @yichengt900, I think we may want to embrace this with or without anammox on.  While the NO3 and O2 depletion condition that it requires to cause a problem is very unlikely, it is possible with anammox off.  @jessluo @yangyzhao5

yichengt900:
@charliestock, no problem. I'll go ahead and update the reference answers.

charliestock:
Thanks @yichengt900, and sorry for the hassle.

PR #156, opened 2025-07-22
neritic_cased_burial allocated to zeros

### LINKED ISSUE #154: `neritic_cased_burial` not allocated to zero
I ran a SPEAR simulation with the latest ocean_BGC code and the latest ESM4.5 configuration (`do_ner_ca_bur = True`), and realized that the variable `neritic_cased_burial` is not allocated to zero before a data_override call is made. See [here](https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/dev/cefi/generic_tracers/generic_COBALT.F90#L4841).

This should work just fine when a xml is properly set up (1. `do_ner_ca_bur = True` in COBALT_override, AND 2. a global on-grid file is provided in the dataTable). In my SPEAR xml, however, I accidentally forgot to add the `neritic_cased_burial` entry line in the dataTable. As data_override() doesn't abort the execution when the variable is not found in the dataTable (and I learned that this is a desired feature), the model continues to use the allocated values, which are garbage, for carbon calculations. My run eventually crashed reporting carbon imbalance (actually global `dic` total goes negative after the first time step) but that doesn't help reveal the true error.

I suggest that we allocate `neritic_cased_burial` to zero to be safe, and perhaps do a check with `SUM(neritic_cased_burial)` after data_override() and report WARNING when `do_ner_ca_bur = True` but `SUM(neritic_cased_burial) = 0`.

@feida6996 @jessluo please ignore what I said earlier today about how data_override() works.

liuxiao37k:
As titled; to make the code more robust.

fixes #154

### DIFF
--- a/generic_tracers/generic_COBALT.F90
+++ b/generic_tracers/generic_COBALT.F90
@@ -4838,7 +4838,7 @@ contains
     ! << Enable neritic CaCO3 burial in shallow regions (depth <= 150m)
     ! Read 'neritic_cased_burial' from netCDF file (O'Mara & Dunne, 2019) to apply spatial pattern
     if (cobalt%do_ner_ca_bur) then
-        allocate(neritic_cased_burial(isd:ied,jsd:jed))
+        allocate(neritic_cased_burial(isd:ied,jsd:jed)); neritic_cased_burial = 0.0
         ! 'neritic_cased_burial' is the 2-D burial field saved in netCDF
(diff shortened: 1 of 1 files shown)


theresa-cordero (approved):

jessluo:
Hi Xiao, thanks for putting together this pull request. This looks good to me too!

PR #157, opened 2025-08-01
Remove diagnoseMLDbyEnergy option

theresa-cordero:
A recent PR to MOM6 [PR 922](https://github.com/NOAA-GFDL/MOM6/pull/922) added a new non-optional argument to `diagnoseMLDbyEnergy`.  This created a problem where `diagnoseMLDbyEnergy` is called in `MOM_generic_tracers.F90`.

We could add this new argument, however that would make ocean_bgc crash with older versions of MOM6. We could require users have an updated version of MOM6, but since the option to use a energy based MLD for the photoacclimation depth was primarily done for completeness, and I think no one is using that option, I suggest we remove it. I have added a fatal error if the option is selected. 

Since these subroutines are within MOM6 and not tested with `MOM_generic_tracer` it seems like a good idea to remove this unneeded dependency.

### DIFF
--- a/generic_tracers/MOM_generic_tracer.F90
+++ b/generic_tracers/MOM_generic_tracer.F90
@@ -774,8 +774,9 @@ subroutine MOM_generic_tracer_column_physics(h_old, h_new, ea, eb, fluxes, Hml, 
       call diagnoseMLDbyDensityDifference(-1, h_old, tv, CS%mld_pha_drho, G, GV, US, CS%diag, &
               CS%mld_pha_href, id_ref_z=-1, id_ref_rho=-1, MLD_out=mld_pha)
     elseif (CS%mld_pha_use_delta_eng) then
-      call diagnoseMLDbyEnergy((/-1, -1, -1/), h_old, tv, G, GV, US, (/CS%mld_pha_deng, &
-              CS%mld_pha_deng, CS%mld_pha_deng/), CS%diag, MLD_out=mld_pha)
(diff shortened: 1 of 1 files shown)


uwagura (approved):
I was able to successfully compile the latest `dev/gfdl` version of the model with these changes

uwagura:
Minor note, but one consequence of this change is that `mld_pha_use_delta_eng` can only have a value of false without crashing the model, so there are now a couple of if statements using this variable that will either never run or only depend on the truth value of the other variable. I guess this also means that it's not really a parameter anymore, since there is only one real choice. We could raise the fatal error during initialization and remove  the `elseif` on line 462 as well, but neither of those changes are necessary, and I guess the current way has the benefit of reminding us this option was removed to avoid the call to `diagnoseMLDbyEnergy`

theresa-cordero:
That's true, I have only added the fatal error here, but we could add it when `mld_pha_use_delta_eng` is read as true from the parameter file. That would at least make the model crash faster for anyone trying to use the unsupported option.

charliestock:
Hello @theresa-cordero.  Sorry for the delayed response on this.  Your approach and the suggested change from @uwagura both sound and look good to me.

theresa-cordero:
I have added an additional error check where the `PHA_MLD_DENG` would be read that will cause the model to crash on initialization if the unsupported option is used.

theresa-cordero:
@charliestock I think this PR is ready to be merged.

PR #161, opened 2025-09-18
Delete unneeded use statements for fm_path_name_len to comply with FMS updates

nikizadehgfdl:
- fm_path_name_len interface is not used in this package but is in use stamements
- FMS2025.04 will remove this interface and references should be removed  in this package to compile with updated FMS
- compiles OK with current and old FMS since fm_path_name_len is unused


theresa-cordero (approved):

theresa-cordero:
Approved. As noted fm_path_name_len is not used in the code.

Issue #158, opened 2025-08-08
Tracer content budget terms were removed


andrew-c-ross:
As we found in the CEFI discussion (https://github.com/NOAA-GFDL/CEFI-regional-MOM6/discussions/224), the BGC tracer content source terms were replaced with the concentration versions by pr #133  (i.e., `jo2` is now equivalent to `jo2c` in older cobalt versions, and the diagnostic named `jo2c` no longer exists). This breaks older budget codes. It also may not be a good idea since for oxygen, the content version of the budget is the first one we got to close. Furthermore, the other content budget terms are still in MOM6 (https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/MOM6/blob/18e8cb3b19bd4d587a06dd27faa0aaed48f7e45b/src/tracer/MOM_tracer_registry.F90#L555), and the concentration terms are still there with the `c` suffix (https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/MOM6/blob/18e8cb3b19bd4d587a06dd27faa0aaed48f7e45b/src/tracer/MOM_tracer_advect.F90#L725).

I propose we consider bringing back some of the content j* terms.

andrew-c-ross:
Solved by merging #165

PR #165, opened 2026-02-24
Diagnostics for tracer content change 

andrew-c-ross:
This PR adds back diagnostic terms for content (concentration * thickness) change due to processes in COBALT as suggested in #158 

There are 8 new diagnostics: jno3h, jnh4h, jpo4h, jsio4h, jndeth, jdich, jalkh, and jo2h. All of these are just the pre-existing concentration terms times the thickness. 

Having the exact thickness being used during the model step is helpful for accurately calculating the budget. Here, I multiply by the thickness that COBALT is seeing (dzt) to ensure it is correct. 

With these diagnostics, the tendency (for oxygen, as an example) is
`o2h_tendency = o2_advection_xy + o2h_tendency_vert_remap + o2_diffusion_xy + o2_vdiffuse_impl + jo2h`

My text editor cleared out a bunch of trailing whitespaces, so the changes in this PR should be viewed with whitespace changes ignored.

### DIFF
--- a/generic_tracers/cobalt_reg_diag.F90
+++ b/generic_tracers/cobalt_reg_diag.F90
@@ -3286,7 +3286,7 @@ module COBALT_reg_diag
     cobalt%id_jalk = register_diag_field(package_name, vardesc_temp%name, axes(1:3),&
          init_time, vardesc_temp%longname,vardesc_temp%units, missing_value = missing_value1)
 
-    vardesc_temp = vardesc("jalkh","Alkalinity content source",'h','L','s','mole eq kg-1 m s-1','f')
+    vardesc_temp = vardesc("jalkh","Alkalinity content source (jalk * thickness)",'h','L','s','mole eq kg-1 m s-1','f')
     cobalt%id_jalkh = register_diag_field(package_name, vardesc_temp%name, axes(1:3),&
          init_time, vardesc_temp%longname,vardesc_temp%units, missing_value = missing_value1)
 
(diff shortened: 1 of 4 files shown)


generic_tracers/cobalt_reg_diag.F90
```diff
@@ -3286,6 +3286,10 @@ subroutine cobalt_reg_diagnostics(diag_list,axes,init_time,phyto,zoo,bact,cobalt
     cobalt%id_jalk = register_diag_field(package_name, vardesc_temp%name, axes(1:3),&
          init_time, vardesc_temp%longname,vardesc_temp%units, missing_value = missing_value1)
 
+    vardesc_temp = vardesc("jalkh","Alkalinity content source",'h','L','s','mole eq kg-1 m s-1','f')
```
charliestock: Is there a clearer description we could put here for the uninitiated?  Maybe just "Alkalinity content source (jnh4*layer thickness)"?  I'm just mention this because I was not fully aware of the purposeful delineation between "content" and "concentration" until this discussion. 

Beyond this PR, I think we may want to eventually add some more explicit direction to users on which variables should be remapped and which should not, but that is a bigger question.

charliestock (commented):
Thanks Andrew.  This looks good.  My only real comment is whether there is a clearer description of the variable names for the uninitiated?  Maybe just "Alkalinity content source (jnh4*the layer thickness)"?  I am torn.

I just mention this because I was not fully aware of the purposeful delineation between "content" and "concentration" until this discussion.

Beyond this PR, I think we may want to eventually add some more explicit direction to users on which variables should be remapped and which should not, but that is a bigger question.

charliestock (approved):
Thanks for making the change Andrew.  This looks good to me.

PR #168, opened 2026-03-17
Move do_vertfill_post into g_tracer_vertdiff_G

### LINKED ISSUE #164: g_tracer_vertfill breaks generic tracer budget diagnostics
I think the call to `g_tracer_vertfill` when the namelist option `do_vertfill_post=.true.` is the source of the perplexing inability to close the near-surface generic tracer budgets even though the column-integrated budgets do close. 

The problem is here:
https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/9c76677aab890c285f5f2cf7b770bfc350493db5/generic_tracers/generic_tracer.F90#L606-L607

Note that the vdiffuse_impl budget diagnostic is calculated within `g_tracer_vertdiff_G`, so after that routine is done and the diagnostic is calculated we are doing an additional rearranging of the tracer content  in `g_tracer_vertfill` that is not included within any budget diagnostic. This especially affects the surface layer for things like oxygen, which is both thin (2 m in z*) and has a sharp concentration gradient due to the surface gas exchange and productivity. The column-integrated budget still closes because the vertfill is conserving the column total. 

To fix this we should probably either move the vertfill into `g_tracer_vertdiff_G` so that it can be included in the vdiffuse_impl diagnostic, or else add a new diagnostic for the change due to vertfill.

andrew-c-ross:
Ensures that the effect of vertfill is included within the vdiffuse_impl budget diagnostic. This should allow budgets to close even with do_vertfill_post enabled.

KD_SMOOTH remains a hard-coded parameter, which is not ideal, but it's probably fine at least until we settle on a solution for reducing noise in vanished layers.


generic_tracers/generic_tracer_utils.F90
```diff
@@ -3447,11 +3449,15 @@ subroutine g_tracer_vertdiff_G(g_tracer, h_old, ea, eb, dt, kg_m2_to_H, m_to_H,
 
     enddo; enddo ! i,j
 
+    ! Do some additional diffusion if requested.
```
charliestock, line 3452: Maybe we can say: "Option to do some additional diffusion to smooth irregularities in vanished layers"? to make the rationale for this clearer?  Should we also indicate whether this is "on" or "off" by default?

charliestock (approved):
Thanks @andrew-c-ross.  I have approved this with some very minor suggestions.  Two other quick questions: 1) Is the default for do_vertfill_post this "true" or "false"? 2) do we also have a do_vertfill_pre still hanging around?

andrew-c-ross:
Thanks @charliestock, yes there is still the vertfill_pre option hanging around: https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/482bf582f0e26d495cda3ffd5e16345c9853d9dc/generic_tracers/generic_COBALT.F90#L3184-L3195 I believe the pre option would also break the budget diagnostics. It is harder to move into a diagnostic term because it sits within `generic_COBALT_update_from_source`, which the biology (eg, jo2h) diagnostic is intended to capture. I would favor removing the vertfill_pre option actually; I'm not aware of any use of it recently (and it also defaults to off). 

I've added clarification that the default for the vertfill_post is off; it's only enabled if do_vertfill_post is explicitly set to true in the input.nml.

andrew-c-ross:
The test passed after the first commit, but it failed after the second commit where the only change was a couple of comments?

andrew-c-ross:
I think we need to update .github/ref/OM4.single_column.COBALT.p4/docker-linux-gnu/ocean.stats now that the CI workflows in CEFI-regional-MOM6 have been updated. That PR was merged between the two commits above.

andrew-c-ross:
#169 should fix the failing test and be merged first

charliestock:
Thanks @andrew-c-ross, I support removing vertfill_pre.  The place to do this, if we were going to do it, would be after the tridiagonal and before the updated tracers are saved.  Lets connect on the second issue at the CEFI code management meeting.

PR #170, opened 2026-04-07
Remove do_vertfill_pre

andrew-c-ross:
Follows up on #168 by removing the namelist option do_vertfill_pre and the logic to apply vertfill when it is true. I'm not aware of any configurations that use this option, and enabling it would break the budget diagnostics.


theresa-cordero:
I approve this PR and can merge it unless you want to wait for @charliestock to provide a review.

charliestock:
Sorry for the delay folks.  Please move ahead with the merge.

uwagura (approved):
This Pr has been verbally approved by several folks, so I'm gonna go ahead an github-approve it so it can by merged in 😄

PR #171, opened 2026-04-08
Write check for do_vertfill_post as two separate ifs

andrew-c-ross:
Fortran does not guarantee short-circuit evaluation, so I should have written this expression as two ifs instead of trying to combine it into one line.

### DIFF
--- a/generic_tracers/generic_tracer_utils.F90
+++ b/generic_tracers/generic_tracer_utils.F90
@@ -3451,8 +3451,10 @@ contains
 
     ! Do some additional diffusion if requested.
     ! This can help smooth irregularities in the vanished layers.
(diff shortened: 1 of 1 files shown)


charliestock (approved):
Thanks @andrew-c-ross.  I guess there was a chance of a crash if the first condition did not result in a short-circuit?  Regardless, it looks good to me!

andrew-c-ross:
Thanks Charlie, yes in theory a segfault or similar could have occurred if `g_tracer_vertdiff_G` was called without the do_vertfill_post argument. Currently the only caller for this subroutine (`generic_tracer_vertdiff_G`) always includes the argument, but better safe than sorry. 

Merging now :rocket:

Issue #167, opened 2026-03-09
COBALT prints excessive warnings in stdout


nikizadehgfdl:
The following code block in generic_COBALT cause far too many warnings printed in stdout increasing the size of logs to a few GB.
These WARNINGS are one per core per timestep which just bloats stdout and makes it unusable.

If this is a serious issue it should be turned into a FATAL. 
Or, we can come up with a way to diagnose the sitatuation at initilization and warn the user there.
Or, we can add up the number of times it occurs during the run and report it at the end of a run (outside main loop) as NOTE.

```
      pha_all_same = all(photo_acc_dpth == photo_acc_dpth(isc,jsc))                                                     
      if (pha_all_same) then                                                                                            
        call mpp_error(WARNING, "Using uniform photoacclimation MLD in COBALTv3 which is not reccomended."//&           
                                "Check that PHA_MLD_CALC is true in the MOM paramter files or you "//&                  
                                "may be using an unrealistic constant value!")                                          
      endif  

```

charliestock:
Thanks for raising this @nikizadehgfdl.  I think your suggestion to diagnose this at initialization and notifying the use there would be perfect, or making it FATAL in the near-term if that is easiest for now.  As the error implies, it is really not a viable option.  Was there a run where you encountered this that may have a bad setting?

Flagging @theresa-cordero so she is aware.

nikizadehgfdl:
I encountered this in a ocean-ice+COBALT experiment that I set up myself. 
The model runs fine for multiple years, although has a huge stdout, so I am not sure that it should be a FATAL.
 
I shall follow what it suggested by adding the following to MOM_override and hopefuilly the messages will go away away.

```
#override PHA_MLD_CALC = True                                                                                           
#override PHA_MLD_USE_DELTA_RHO = True                                                                                  
#override PHA_MLD_HREF = 10.0 ! can also remove for the default of 0.0                                                  
#override PHA_MLD_DRHO = 0.03 
```

charliestock:
That sounds good Niki.  Let's leave this issue open as well.  It would be good to discuss how to limit the standard output in sensible ways to prevent huge files.  The "is_root_pe" suggested by @andrew-c-ross could work, but doing this check at initialization may  be the be the best long-term solution.

nikizadehgfdl:
We agreed in a DOC&DEV meeting to convert these "WARNING" to "FATAL"

Issue #166, opened 2026-03-09
generic_tracer_utils prints excessive spurious warnings in the stdout


nikizadehgfdl:
The following block cause too many warnings (one per core) to be printed to stdout causing stdout to grow to GBs.
Given that there are already NOTEs printed for to make users aware of these issues in field_table, these warnings are not required and waste compute/storage resources. 
I suggest we remove this block iof code   from generic_tracer_utils:

```
 if(errorstring .ne. '') then                                                                                        
       !The following cannot be FATAL for backward compatibility with MOM5 and GOLD                                     
       call mpp_error(WARNING, trim(sub_name) // ' : there are tracers with required source properties that are not set '//&
                      'in the field_table. Grep the stdout for NOTEs from g_tracer_print_info and correct the field_table!')
    endif                                                                                                               
 ```

**Comment by charliestock** [comment]
Thanks @nikizadehgfdl.  We could remove, though it is a great example of an error message that would help a user diagnose an issue.  Can we now make it FATAL?  Is Backward compatability from MOM5 to GOLD still an issue?

**Comment by andrew-c-ross** [comment]
Can we wrap the call to `g_tracer_print_info` with an `if (is_root_pe())`? There's no reason the properties of g_tracer should be different across PEs right? 

https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/482bf582f0e26d495cda3ffd5e16345c9853d9dc/generic_tracers/generic_tracer.F90#L216

**Comment by nikizadehgfdl** [comment]
The true content of this WARNING message already appears as NOTEs (as seen below), and this warning is just redundant and confusing , asking the user to look for those NOTEs in the stdout. 
It should not be FATAL since the model runs regardless.
The comment regarding MOM5 and GOLD are not relevant anymore. 
I think this block could safely be removed.  

E.g., the NOTEs are already there in stdout. 
```
NOTE from PE     0: g_tracer_print_info: femd : src_var_gridspec is not set in the field_table
NOTE from PE     0: g_tracer_print_info: femd : src_var_gridspec is not set in the field_table
NOTE from PE     0: g_tracer_print_info: femd : src_var_gridspec is not set in the field_table
NOTE from PE     0: g_tracer_print_info: femd : src_var_gridspec is not set in the field_table

```

nikizadehgfdl:
There was an agreement to remove these lines in a DOC&DEV meeting.

PR #173, opened 2026-04-15
Fixes issues #166 and #167 for excessive stdout warnings

nikizadehgfdl:
- Closes issue #166, the printing of missing source entries in field_table is repeated and unnecessary
- Closes issue #167, Using uniform photoacclimation MLD should stop the run at the onset so the user can correct it


theresa-cordero (approved):

theresa-cordero:
I ran a test with NWA12 to confirm this would not cause a crash and it works as expected.

PR #174, opened 2026-05-11
Remove unused EOS subroutines

uwagura:
Short cleanup PR that simply removes unused EOS imports from `generic_Cobalt.F90` and `generic_tracer.F90` so that COBALT has fewer dependencies on MOM. This will hopefully prevent either of these files from recompiling when there are changes in MOM affecting MOM_EOS.F90. 

As a side note, there are other imports in `MOM_generic_tracers.F90` that don't appear to be used either. I did not include them in this PR since they are all parts of modules that are used, so removing them probably won't help prevent recompiles, but I thought I'd mention it anyways; 
```
use MOM_coms, only : EFP_type, max_across_PEs, min_across_PEs, PE_here ! last three unused
use MOM_diagnose_mld,  only : diagnoseMLDbyDensityDifference, diagnoseMLDbyEnergy ! last one unused
use MOM_diag_mediator, only : post_data, register_diag_field, safe_alloc_ptr ! none used
use MOM_diag_mediator, only : diag_ctrl, get_diag_time_end
use MOM_error_handler, only : MOM_error, FATAL, WARNING, NOTE, is_root_pe ! last unused
...
use MOM_sponge, only : set_up_sponge_field, sponge_CS ! first unused
use MOM_time_manager, only : time_type, set_time
use MOM_tracer_diabatic, only : tracer_vertdiff, applyTracerBoundaryFluxesInOut ! first unused
```

### DIFF
--- a/generic_tracers/MOM_generic_tracer.F90
+++ b/generic_tracers/MOM_generic_tracer.F90
@@ -33,9 +33,8 @@ use MOM_ALE_sponge, only : set_up_ALE_sponge_field, ALE_sponge_CS
 use MOM_ALE_sponge, only : ALE_sponge_CS, initialize_ALE_sponge
 use MOM_coms, only : EFP_type, max_across_PEs, min_across_PEs, PE_here
 use MOM_diagnose_mld,  only : diagnoseMLDbyDensityDifference, diagnoseMLDbyEnergy
-use MOM_diag_mediator, only : post_data, register_diag_field, safe_alloc_ptr
 use MOM_diag_mediator, only : diag_ctrl, get_diag_time_end
-use MOM_error_handler, only : MOM_error, FATAL, WARNING, NOTE, is_root_pe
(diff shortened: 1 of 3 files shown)


theresa-cordero:
Could you add a comment here or open an issue with the list of other unused imports? I think it is best to remove all unneeded imports.

theresa-cordero (approved):

uwagura:
Since I know where most of the other imports are / can easily find the others, do you think it would better to just convert this PR into a "remove unused imports" PR by adding commits to clean up the other imports

uwagura:
@theresa-cordero I cleaned up some of the other imports in `MOM_generic_tracer.F90`. I took a quick look at the files, and I didn't see anything that I thought needed to be removed now, thought I mostly skimmed them. This branch compiles and runs the 1d case, so if everything looks good to you we can merge it in.

theresa-cordero (approved):
Approved!

PR #178, opened 2026-07-07
Simplify bottom layer diagnostics routines into one new, small module

andrew-c-ross:
This PR refactors the calculation of bottom diagnostics for generic tracers in COBALT into a new module, `generic_bottom_layer_diags`. 

The benefits of doing this are
1. Less repeated code
2. Easier to add new bottom diagnostics (just register, call generic_bld_average, and send)
3. Potentially simpler to change how the bottom averages are calculated in the future

The algorithm used for bottom averaging here is an exact duplicate of the old code (averaging over `bottom_thickness`) so the diagnostic output should not change.

AI use: these changes were planned and reviewed with help from Opus 4.8, but the implementation and code were fully done by hand.

### DIFF
--- a/generic_tracers/cobalt_send_diag.F90
+++ b/generic_tracers/cobalt_send_diag.F90
@@ -296,8 +296,6 @@ module COBALT_send_diag
             model_time, rmask = grid_tmask(:,:,1), is_in=isc, js_in=jsc,ie_in=iec, je_in=jec)
 
           ! Calculate bottom averaged diagnostics.
-          ! Note that averages for btm_o2, btm_no3, btm_co3_ion, and btm_co3_sol_calc
-          ! were already calculated in generic_COBALT_update_from_source.
           call generic_bld_average(cobalt%bld, cobalt%p_alk(:,:,:,tau), cobalt%btm_alk)
           call generic_bld_average(cobalt%bld, cobalt%p_dic(:,:,:,tau), cobalt%btm_dic)
           call generic_bld_average(cobalt%bld, Temp(:,:,:), cobalt%btm_temp)
@@ -311,6 +309,12 @@ module COBALT_send_diag
             call generic_bld_average(cobalt%bld, cobalt%f_co3_ion, cobalt%btm_co3_ion)
             call generic_bld_average(cobalt%bld, cobalt%co3_sol_calc, cobalt%btm_co3_sol_calc)
           endif
+
+          ! btm_o2 and btm_no3 were calculated during the update from source, 
+          ! but we always need to recalculate them here because their fields are changed
+          ! later in that routine.
+          call generic_bld_average(cobalt%bld, cobalt%p_o2(:,:,:,tau), cobalt%btm_o2)
+          call generic_bld_average(cobalt%bld, cobalt%p_no3(:,:,:,tau), cobalt%btm_no3)
 
           do j = jsc, jec ; do i = isc, iec
             cobalt%btm_omega_calc(i,j) = 0.0
--- a/generic_tracers/cobalt_send_diag.F90
+++ b/generic_tracers/cobalt_send_diag.F90
@@ -50,7 +50,6 @@ module COBALT_send_diag
       real :: drho_dzt
       real, dimension(:,:,:) ,pointer :: grid_tmask
       integer, dimension(:,:),pointer :: mask_coast,grid_kmt
-      integer, dimension(:,:), Allocatable :: k_bot
       real, dimension(:,:), Allocatable :: rho_dzt_100,rho_dzt_200
       integer :: k_100,k_200
       real, dimension(:,:), Allocatable :: field_2d !used to calculate some 2d fields before saving
--- a/generic_tracers/cobalt_send_diag.F90
+++ b/generic_tracers/cobalt_send_diag.F90
@@ -20,6 +20,7 @@ module COBALT_send_diag
 
   use g_tracer_utils, only : g_send_data,g_tracer_get_pointer,g_tracer_set_values
   use g_tracer_utils, only : g_tracer_type, g_tracer_get_common
+  use generic_bottom_layer_diags, only : generic_bld, generic_bld_average
   use FMS_co2calc_mod, only : FMS_co2calc, CO2_dope_vector
 
   implicit none; private
@@ -50,7 +51,7 @@ module COBALT_send_diag
       real, dimension(:,:,:) ,pointer :: grid_tmask
       integer, dimension(:,:),pointer :: mask_coast,grid_kmt
       integer, dimension(:,:), Allocatable :: k_bot
-      real, dimension(:,:), Allocatable :: rho_dzt_100,rho_dzt_200,rho_dzt_bot
+      real, dimension(:,:), Allocatable :: rho_dzt_100,rho_dzt_200
       integer :: k_100,k_200
       real, dimension(:,:), Allocatable :: field_2d !used to calculate some 2d fields before saving
       real, dimension(:,:,:), Allocatable :: flux_i !used to save fluxes at the interfaces
@@ -187,6 +188,9 @@ module COBALT_send_diag
               cobalt%nmd_misc(i,j,k)=phyto(MEDIUM)%f_n(i,j,k) - phyto(MEDIUM)%f_n(i,j,k)*phyto(MEDIUM)%silim(i,j,k)
             enddo; enddo ; enddo !} i,j,k
 
+            call generic_bld_average(cobalt%bld, cobalt%f_co3_ion, cobalt%btm_co3_ion)
+            call generic_bld_average(cobalt%bld, cobalt%co3_sol_calc, cobalt%btm_co3_sol_calc)
+
           endif !} recalculate carbon system properties
 
           used = g_send_data(cobalt%id_co3_sol_arag, cobalt%co3_sol_arag, &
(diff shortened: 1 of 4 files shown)


generic_tracers/generic_bottom_layer_diags.F90
```diff
@@ -0,0 +1,91 @@
...
+
+  public generic_bld_alloc, generic_bld_dealloc, generic_bld_update, generic_bld_average
+
+  contains
+
+    subroutine generic_bld_alloc(bld, isc, iec, jsc, jec, Rho_0, bottom_thickness)
+      type(generic_bld), intent(inout) :: bld
+      integer, intent(in) :: isc, iec, jsc, jec
+      real, intent(in) :: Rho_0, bottom_thickness
+      allocate(bld%kmt(isc:iec, jsc:jec))
```
uwagura: Does the value of kmt ever depend on `(i,j)`? It looks like to me it's hard coded to the same value for every `(i,j)`:

https://github.com/NOAA-CEFI-Regional-Ocean-Modeling/ocean_BGC/blob/5760648c59983b66aa44591b1a6a0133f852f333/generic_tracers/MOM_generic_tracer.F90#L479-L484

Unless I'm missing something it looks like we can just replace this array with a scalar? ( though this is more of a cobalt wide change that could be part of a different PR) 

Regarding this PR specifically: it looks like the bld subroutine are only called within subroutines where kmt is already defined (  cobalt_send_diags, generic_COBALT_update_from_source). Instead of storing a copy of kmt in the generic_bld type, could we just pass it into the subroutines that need it?

generic_tracers/cobalt_send_diag.F90
```diff
@@ -50,7 +51,7 @@ subroutine cobalt_send_diagnostics(tracer_list,model_time,grid_tmask,Temp,Salt,r
       real, dimension(:,:,:) ,pointer :: grid_tmask
       integer, dimension(:,:),pointer :: mask_coast,grid_kmt
       integer, dimension(:,:), Allocatable :: k_bot
```
uwagura: I don't think k_bot is used anymore in cobalt_send_diag after this pr, so it can be be removed from the argument list

generic_tracers/generic_bottom_layer_diags.F90
andrew-c-ross: I think you're right that grid_kmt is the same everywhere. Maybe this is leftover from older MOM4/5? 

My philosophy for keeping a copy of kmt was to minimize as much as possible the amount of boilerplate. I think the other data required by the averaging function are also available, but I wanted to be able to do `generic_bld_average(bld, field, field_btm)` instead of `generic_bld_average(isc, iec, jsc, jec, Rho_0, bottom_thickness, rho_dzt, grid_kmt, field, field_btm)`

andrew-c-ross: In hindsight it would probably make more sense to have kmt be a pointer just like rh_dzt instead of a copy.

generic_tracers/generic_bottom_layer_diags.F90
```diff
@@ -0,0 +1,91 @@
...
+        bld%k_bot(i,j) = 0
+        bld%rho_dzt_bot(i,j) = 0.0
+        if (grid_kmt(i,j) .gt. 0) then
+          do k = grid_kmt(i,j),1,-1
+            ! Check if the top of layer k is within the bottom thickness.  If so, include its properties in the bottom
+            ! layer averages.  Overshoots will be subtracted off later.
+            if (bld%rho_dzt_bot(i,j).lt.(bld%Rho_0*bld%bottom_thickness)) then
+              bld%k_bot(i,j) = k
+              bld%rho_dzt_bot(i,j) = bld%rho_dzt_bot(i,j) + bld%rho_dzt(i,j,k)
+            endif
```
uwagura, line 69: Once this condition is no longer true, it won't be true for any layer above `k`. Perhaps we could save a few iterations by breaking the loop early in those instances?:
```suggestion
            else
              exit
            endif
```
(edit: I'm not sure if that is the correct way to break out of a nested loop in fortran but hopefully you get the idea)

generic_tracers/generic_bottom_layer_diags.F90
uwagura: Making grid_kmt a scalar would simplify code throughout cobalt. @charliestock , @jessluo , do either of you have any thoughts on this?

jessluo: I believe `grid_kmt` is zero over land... so then we will need some way to distinguish between land and ocean points.

andrew-c-ross: `G%mask2dT` should be all that's needed for the land mask

jessluo: yes, the usage of `grid_kmt` is a bit confusing, as in some places it's used as a land mask and in other places it's used for just indicating the bottom layer of the model (a scalar, typically 75). I do think it'll be good to separate the usages into specific variables fit for each purpose but perhaps it should be in a separate PR, as that change will touch many other parts of the generic_tracers repository?

andrew-c-ross:
I should add checks for whether the pointers are associated, and informative errors if not, to `generic_bld_average`.

jessluo (approved):
This overall looks good to me and seems like a nice improvement over the ad-hoc way we were previously calculating the bottom concentrations. 

Eventually, I would like the flexibility for `bottom_thickness` to be spatially variable, i.e., become a 2-D array, so that it could functionally represent the thickness of a benthic boundary layer, but that would be a much larger project involving folks working on the ocean model. And I suspect it would not be calculated here in `generic_tracers` but more so on the MOM side. 

If you can see a reasonable path forward for this variable to become a 2-D array, then we can leave it now as is (a scalar) and just keep in mind that eventually we will probably want it to become an array.

generic_tracers/cobalt_send_diag.F90
```diff
@@ -295,73 +298,29 @@ subroutine cobalt_send_diagnostics(tracer_list,model_time,grid_tmask,Temp,Salt,r
...
-            cobalt%btm_co3_sol_arag(i,j) = 0.0
-            cobalt%btm_co3_sol_calc(i,j) = 0.0
-            cobalt%btm_co3_ion(i,j) = 0.0
+          call generic_bld_average(cobalt%bld, cobalt%p_o2(:,:,:,tau), cobalt%btm_o2)
+          call generic_bld_average(cobalt%bld, cobalt%p_alk(:,:,:,tau), cobalt%btm_alk)
+          call generic_bld_average(cobalt%bld, cobalt%p_dic(:,:,:,tau), cobalt%btm_dic)
+          call generic_bld_average(cobalt%bld, Temp(:,:,:), cobalt%btm_temp)
+          call generic_bld_average(cobalt%bld, cobalt%f_htotal(:,:,:), cobalt%btm_htotal)
+          call generic_bld_average(cobalt%bld, cobalt%co3_sol_arag(:,:,:), cobalt%btm_co3_sol_arag)
+
```
rakshit-s, line 306: Just curious, why the `cobalt%btm_co3_sol_arag` is calculated here, but `cobalt%btm_co3_sol_calc` in a previous section within "recalculate carbon system properties"? 
Can `cobalt%btm_co3_sol_calc` and also `cobalt%btm_co3_ion` be moved here so that all the `call generic_bld_average` are at the same place in cobalt_send_diag.F90 file?

```suggestion
          call generic_bld_average(cobalt%bld, cobalt%f_co3_ion(:,:,:), cobalt%btm_co3_ion)
          call generic_bld_average(cobalt%bld, cobalt%co3_sol_calc(:,:,:), cobalt%btm_co3_sol_calc)
          
```

generic_tracers/cobalt_send_diag.F90
```diff
@@ -187,6 +187,9 @@ subroutine cobalt_send_diagnostics(tracer_list,model_time,grid_tmask,Temp,Salt,r
               cobalt%nmd_misc(i,j,k)=phyto(MEDIUM)%f_n(i,j,k) - phyto(MEDIUM)%f_n(i,j,k)*phyto(MEDIUM)%silim(i,j,k)
             enddo; enddo ; enddo !} i,j,k
 
+            call generic_bld_average(cobalt%bld, cobalt%f_co3_ion, cobalt%btm_co3_ion)
+            call generic_bld_average(cobalt%bld, cobalt%co3_sol_calc, cobalt%btm_co3_sol_calc)
+
```
rakshit-s: In case it is moved further down in the code with other `call generic_bld_average`
```suggestion

```

rakshit-s (commented):
Suggesting moving all `call generic_bld_average' at a same place in cobalt_send_diag.F90.

rakshit-s (approved):
I went through all the changes and it all looks good to me. There is one minor suggestion to reorganize the 'call generic_bld_average', in a previous comment. Approving this review.
This might be a good opportunity to add other btm conc terms that would be needed for the CBED module. Terms to add (through a future PR) `btm_nh4`, `btm_po4`, `btm_sio4`, `btm_fed`.

andrew-c-ross:
@rakshit-s good point about the organization. I moved the co3 ion and sol_calc calculations down to where the other diagnostics are, along with more comments explaining that these diagnostics are also calculated in generic_COBALT_update_from_source.

andrew-c-ross:
This PR actually changed the diagnostic output for btm_o2 because it removed a recalculation of btm_o2 after updating from source and before sending diags that the old code had. I added that back in now, and the btm_o2 diagnostic is identical in my loop current test. 

However, I noticed that even before this PR, btm_no3 was calculated early in the update from source and not re-calculated after no3 was modified. I added that in to this PR now, so it will change diagnostics for btm_no3 (but not the actual model no3). I think the new answers are more correct, but I'd appreciate maybe a quick check from @charliestock or @rakshit-s to confirm.

Aside from that, the other btm_ diagnostics were identical to older results in my loop current test.

andrew-c-ross:
Just talked with Charlie about the btm_no3 and we agreed it should be recalculated before sending the diagnostic, and that it probably wasn't in any diag tables so it should be fine to change answers for. I'm going to merge this  🚀

PR #179, opened 2026-07-07
Delete unused variables and cleanup whitespace

uwagura:
As titled. This is a simple PR to remove some unused variables from `FMS_co2calc.F90` and `generic_COBALT`, and  remove an extra call to `g_tracer_get_common` in `cobalt_send_diag.F90`. In the process it also cleans up some white space in generic_COBALT.F90. This is just a smattering if issues I noticed while looking through dev/cefi, which is why the PR touches such random parts of the model. If folks notice other minor cleanup they would like to add to this PR, feel free to do so.


andrew-c-ross:
With `htotallo` and `htotalhi` no longer being set to anything in generic COBALT in this PR, is it worth completely eliminating all of the allocate/deallocates and the inclusions in the calls to `FMS_co2calc`? The latter should be possible by naming the argument for htotal (`call FMS_co2calc(..., htotal=cobalt%f_htotal(:,:,k), ...)`)

generic_tracers/generic_COBALT.F90
```diff
@@ -3042,9 +3038,6 @@ end subroutine generic_COBALT_update_from_bottom
   !   This is the subroutine to contain most of the biogeochemistry for calculating the
   !   interaction of tracers with each other and with outside forcings.
   !  </DESCRIPTION>
-  !  <TEMPLATE>
-  !   call generic_COBALT_update_from_source(tracer_list,Temp,Salt,dzt,hblt_depth,&
-  !                                         ilb,jlb,tau,dt, grid_dat,sw_pen,opacity)
   !  </TEMPLATE>
```
andrew-c-ross: This stray closing tag can be deleted too

jessluo:
> With `htotallo` and `htotalhi` no longer being set to anything in generic COBALT in this PR, is it worth completely eliminating all of the allocate/deallocates and the inclusions in the calls to `FMS_co2calc`? The latter should be possible by naming the argument for htotal (`call FMS_co2calc(..., htotal=cobalt%f_htotal(:,:,k), ...)`)

Yes, I would support this change!

generic_tracers/generic_COBALT.F90
uwagura: hmm I don't actually recall why I wanted to get rid of this template - I'm guessing it's just because it's super out of date?

I'm wondering if I should just leave this comment alone for now, since the whole thing is out of date anyways and may be worth updating as part of a documentation cleanup PR at a later date. On that note, do we have an active doxygen site for COBALT that renders any of these comments?

andrew-c-ross: I'm ok with whatever is easiest. I think this is a very old GFDL custom documentation format. Moving to doxygen or something else more modern would be a good doc&dev idea.

andrew-c-ross: The coupler seems to be moving from doxygen to FORD: https://github.com/NOAA-GFDL/FMScoupler/issues/170

generic_tracers/cobalt_types.F90
jessluo: also will need to remove these variables too, or at least htotal_scale_lo and htotal_scale_hi

generic_tracers/generic_COBALT.F90
jessluo: should remove these calls to `htotal_scale_lo` and `htotal_scale_hi` below

jessluo (changes requested):
looks good, the main additional removal would be the `htotal_scale_{lo,hi}` variables.

generic_tracers/generic_COBALT.F90
uwagura: I'm going to leave this as is for now, and we can revisit the documentation for this subroutine as part of a separate change

andrew-c-ross: I've been trying out FORD. Seems like it might do everything we need. I'll see if I can put in a PR to make it build docs as an action, and then we could start converting regular comments to actual documentation

andrew-c-ross (approved):

jessluo (approved):

PR #180, opened 2026-07-07
Explicitly deallocate and reduce size of arrays used in source/sink imbalance calc and for storing thicknesses

uwagura:
This PR does several things: 

1.) It adds explicit `deallocate` statements for arrays used to store thicknesses and calculate imbalances after the source/sink calculations

2.) It reduces `ztop` from `3d` to `2d` since we don't need to keep track of previous k values of `ztop` in a given layer

3.) It converts the 3d `post_tot<nutrient>` arrays to scalars, since the values those arrays hold are only used once for element-wise checks. 

The original intent of this PR was just to be explicit about de-allocating allocated memory - I think the [compiler technically handles](https://www.fortran90.org/src/best-practices.html#allocatable-arrays) this for us, but I thought it would be cleaner to include those statement. I realized in the process that most of these arrays could be made smaller, which has the added benefit of getting rid of the expensive 3D allocations that come with each`post_tot` array in each call to `generic_COBALT_update_from_source`. 

This seemed to improve the runtime of the NEP10.COBALT test case slightly, though to be honest I can't find my reference times on c5 and the difference isn't huge as COBALT is not the most expensive part of that test case. Still, this approach seems a little more memory friendly, so I figured I would open the PR.

### DIFF
--- a/generic_tracers/generic_COBALT.F90
+++ b/generic_tracers/generic_COBALT.F90
@@ -3166,12 +3166,13 @@ contains
     real, dimension(:,:),   Allocatable :: neritic_cased_burial
     ! >>
 
-    real, dimension(:,:,:), Allocatable :: ztop, zmid, zbot
     real, dimension(:,:,:), Allocatable :: pre_totn, net_srcn, post_totn
     real, dimension(:,:,:), Allocatable :: pre_totp, net_srcp, post_totp
     real, dimension(:,:,:), Allocatable :: pre_totsi, post_totsi
     real, dimension(:,:,:), Allocatable :: pre_totfe, net_srcfe, post_totfe
     real, dimension(:,:,:), Allocatable :: pre_totc, net_srcc, post_totc
+    real, dimension(:,:),   Allocatable :: ztop
+    real, dimension(:,:,:), Allocatable :: zmid, zbot
     real, dimension(:,:),   Allocatable :: pka_nh3,phos_nh3_exchange
 
     real :: tr,ltr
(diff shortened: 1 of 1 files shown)


generic_tracers/generic_COBALT.F90
```diff
@@ -6074,29 +6080,28 @@ subroutine generic_COBALT_update_from_source(tracer_list,Temp,Salt,rho_dzt,dzt,h
     ! day-1, so an imbalance of order 1 would be very large whereas 1e-9 is very small.
     ! A reccomended tolerance is between 1e-7 and 1e-9.
     imbal_flag = 0;
-    stdoutunit = stdout();
```
andrew-c-ross: With this gone, we can delete line 3181 `integer :: stdoutunit, imbal_flag, outunit` provided `imbal_flag = 0;` on line 6082 is also removed (it is never used).

andrew-c-ross:
FWIW claude also suggests

> 
>   zbot is bitwise equal to cobalt%zt. Both start at dzt(i,j,1). Both use the recurrence x(k) = x(k-1) + dzt(k). cobalt%zt is written only at lines 3211 and 3218. The zt= argument at lines 3248 and 3272 is an optional input to FMS_co2calc, so no call modifies it. Delete zbot. Use cobalt%zt(i,j,k) at line 4974. This removes a full 3D array.
> 
>   zmid is a 3D array with one useful slice. The only read is zmid(i,j,nk) at line 3650. Replace it with cobalt%zt(i,j,nk-1) + 0.5*dzt(i,j,nk), which is bit-identical. This form assumes nk >= 2. Line 4580 already uses cobalt%zt(i,j,nk) for a similar depth test.
> 
>   After both changes, ztop has no purpose. The whole block at lines 3207-3224 disappears. The saving is two more nx*ny*nk arrays, on top of what the PR already saves.

generic_tracers/generic_COBALT.F90
andrew-c-ross: Nevermind, looks like that was done in #179.

uwagura:
@jessluo @charliestock , I have removed ztop and zbot from this PR as suggested by Claude and as discussed at the last doc and dev meeting. I left zmid in the code, I just made it a 2d a layer since it is only read at a single layer and renamed it to reflect this fact. Claude suggested getting rid of this array altogether, but I think the check involving zmid  is a bit more legible with an extra array rather than trying to calculate it within the if statement. Let me know what you think of this. Otherwise, I think this PR is good to go.

andrew-c-ross (approved):
This compiled and ran in the loop current test case successfully, and the diagnostic outputs matched results from the current dev/cefi, except for btm_no3 which is expected because dev/cefi has #178 merged.

